PR #180, opened 2015-01-16
DLTP-329 Further refining print styles

danhorst:
The Galaxie Polaris Condensed webfont we are using has a printing
artifact with a capital C on Xerox printers; reverting to another
typeface.

Adding page margins.


cdeluca495:
That's pretty obscure!  :+1:

PR #187, opened 2015-02-10
Department name updates

danhorst:
Pending review by @cdeluca495


cdeluca495:
That there is a lot of ands. :+1:

PR #239, opened 2015-10-19
Increasing no items found text

### LINKED ISSUE #221: Clearly identify/communicate when there is nothing available
_Given a search that returns no results_

From [Reid’s recommendations](https://docs.google.com/a/nd.edu/document/d/1lCbb_AUX6cFjya1cHcr355HKu95-7NfCgvQ1Jw-sofA/edit?usp=sharing):
> The “no entries found” message needs to be clearly displayed so the user does not miss it or continue to wait for records to appear.

> - Move this response closer to the center (where the user would more naturally be looking for results)
> - Make the text bigger and/or darker
> - OUT OF SCOPE ~~Eventually: offer a link to advanced search component and/or help with constructing a search (this might even be a pop-up box instead of an entire page).~~

---
[Trello card](https://trello.com/c/uEhZl87g)

jeremyf:
Why:

* At present the "No entries found" was getting lost in the chatter of
  the page.

This change addresses the need by:

* Increasing the font-size and font-weight of the "no search items"
  result. Note: This value piggy backs on the paging response for a
  search query (and not the search result area). In other words, if
  there are no items in the result set; Instead of rendering a "no
  items in the result set" section within the search results area,
  Blacklight renders a "no_items_found" I18n value in the pagination
  area.

As per the ticket, I'd like to move the result to the left, however,
due to the parent container's lack of a modifying class, I can' remove
or as easily target and adjust the parent `.pager { float: right; }`
directive.

Closes #221

### DIFF
--- a/app/assets/stylesheets/modules/search_results.css.scss
+++ b/app/assets/stylesheets/modules/search_results.css.scss
@@ -1,12 +1,5 @@
 @import 'global-variables';
 
-.page_entries {
-  .no-items-found {
-    font-size: 1.33em;
-    font-weight: bold;
-  }
-}
-
 .search-results-list {
   list-style-type: none;
   margin: 0;
--- a/app/assets/stylesheets/modules/search_results.css.scss
+++ b/app/assets/stylesheets/modules/search_results.css.scss
@@ -1,5 +1,12 @@
 @import 'global-variables';
 
+.page_entries {
+  .no-items-found {
+    font-size: 1.33em;
+    font-weight: bold;
+  }
+}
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/modules/search_results.css.scss
```diff
@@ -1,5 +1,12 @@
 @import 'global-variables';
 
+.page_entries {
```
jeremyf: Except that this is a Blacklight generated HTML class. So no changes will be made.

PR #251, opened 2015-10-26
If a DOI does not have a prefix pass it to Plumx

danhorst:
More view logic! :-1:

### DIFF
--- a/app/views/curation_concern/base/show.html.erb
+++ b/app/views/curation_concern/base/show.html.erb
@@ -55,7 +55,12 @@
 
   if has_doi
     doi = curation_concern.identifier
-    unprefixed_doi = doi.split(':').drop(1).join(':')
+    split_doi = doi.split(':')
+    if split_doi.length > 1
+      unprefixed_doi = split_doi.drop(1).join(':')
+    else
+      unprefixed_doi = doi
+    end
   end
 %>
 <% if has_doi %>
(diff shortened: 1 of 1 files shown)


app/views/curation_concern/base/show.html.erb
```diff
@@ -55,7 +55,12 @@
 
   if has_doi
     doi = curation_concern.identifier
-    unprefixed_doi = doi.split(':').drop(1).join(':')
+    split_doi = doi.split(':')
+    if split_doi.length > 1
```
dbrower: what kind of strings are you protecting against here?

Would it make sense to test the prefix of the string for "doi:" specifically?

danhorst: Yes, an explicit check would be fine. Plum requires the DOI to not start with 'doiI:'.

dbrower: I was wondering if it started with an http: (which is non-conforming to the extreme). also it might be uppercased, `DOI:`.

danhorst: For this issue filtering addional protocols or URN prefixes is out of scope. I could check via regex instead e.g. /DOI:/i

It does make sense to have a parser validate user input for DOIs because we allow people to just paste in whatever. Want to take a crack at writing/finding one @dbrower?

dbrower: ha. maybe? this is just asking for trouble.

PR #258, opened 2015-10-28
Consolidate and lint JavaScript files

danhorst:
Linting and rearranging JavaScript files in CurateND.
This sets the stage for more dramatic refactoring later.


danhorst:
@jeremyf let's hold on merging this until after the current release.

I don't want to derail our schedule if I accidently broke something.

Issue #257, opened 2015-10-28
Add Login button to 401 pages


dbrower:
Only display the button if the current user is not already logged in. Add the button to each of the three unauthorized pages (are three separate pages needed?):

 * https://github.com/ndlib/curate_nd/blob/master/app/views/errors/401.html.erb
 * https://github.com/ndlib/curate_nd/blob/master/app/views/curation_concern/base/unauthorized.html.erb
 * https://github.com/ndlib/curate_nd/blob/master/app/views/curate/collections/unauthorized.html.erb

PR #260, opened 2015-10-28
Add login button option for 401 pages.

dbrower:
Also change unauthorized page to display the noid instead of the item
id (which includes the fedora namespace).

issue #257

### DIFF
--- a/app/views/curate/collections/unauthorized.html.erb
+++ b/app/views/curate/collections/unauthorized.html.erb
@@ -1,3 +1,7 @@
 <h1>Unauthorized</h1>
-<p>The collection you have tried to access is private<p>
-<p>ID: <%= @collection.id %>
+<p>The collection you have tried to access is private.</p>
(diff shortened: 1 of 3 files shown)


app/views/curate/collections/unauthorized.html.erb
```diff
@@ -1,3 +1,7 @@
 <h1>Unauthorized</h1>
-<p>The collection you have tried to access is private<p>
-<p>ID: <%= @collection.id %>
+<p>The collection you have tried to access is private.</p>
+<p>ID: <%= @collection.id %></p>
```
danhorst: Should this be `.noid` as well?

dbrower: Probably. I noticed it too, but didn't change it since I wasn't sure whether a collection object had a #noid method and I was on my way out.

jeremyf:
I added another issue so we can move forward with the deploy. See #261

PR #265, opened 2015-10-30
Change collection id to noid in unauthorized page

### LINKED ISSUE #261: Switch from ID to NOID on collection unauthorized page
In the given view `app/views/curate/collections/unauthorized.html.erb`, see if we can switch from `@collection.id` to `@collection.noid`.

dbrower:
fixes #261


jeremyf:
:+1:

PR #272, opened 2015-11-02
Stakeholder requested changes

danhorst:
This is a batch of several content changes that came out of the maintenance sprint closeout meetings.


jeremyf:
:+1:

PR #276, opened 2015-11-04
Forcing ETDs to be rendered through the common objects viewer

danhorst:
Several ETDs do not render correctly via the show action under curation concerns. This change bypasses the problem rather instead of fixing it directly.

Here are some examples:
https://curate.nd.edu/concern/etds/4f16c24994w
https://curate.nd.edu/show/4f16c24994w
https://curate.nd.edu/concern/etds/w3763487g4k
https://curate.nd.edu/show/w3763487g4k

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,10 @@ CurateNd::Application.routes.draw do
 
   root 'static_pages#home'
 
+  # Some ETDs are not loading correctly on the curation concern page
+  get '/concern/etds/:id', to: redirect { |params, request|
+    "/show/#{params[:id]}"
+  }
   curate_for
 
   namespace :admin do


jeremyf:
Is this an acceptable solution? Who decides if its acceptable?

dbrower:
What is the original view problem? I don't notice it when I click the links in the OP.

danhorst:
@jeremyf this is a stop-gap solution.
@dbrower the selected ETDs are not visible to campus through the curation concern renderer. They are visible to campus through the common objects controller (as their access controls suggest they should be).

The problem warrants some investigation to see which assumptions about the legacy ETD data are tripping up the application. However, I don’t want this to be a time sink as we are planning to move _all_ curation concern rendering to the forthcoming replacement for the common objects viewer.

@jeremyf please take a look at this but cap it at 2 hours.

jeremyf:
My gut tells me I won't get very far with 2 hours. So I'm okay with this being a stop-gap solution.

PR #286, opened 2015-11-16
The thumbnail partial also makes a link

dbrower:
Also removes a hardcoded pid from the view.


dbrower:
@danhorst this is to point out a problem. I didn't look for the same problem elsewhere, or verify that the link made in the thumbnail has the correct styles.

danhorst:
Looks like I committed that by mistake. I don't have thumbnail generation running on my dev machine so I used an existing item to troubleshoot layout issues.

PR #289, opened 2015-11-17
Download error page

danhorst:
As pointed out by @cdeluca495 the downloads controller does not benefit from the error handling present in much of the rest of the application.

I added a simple capture for Hydra::AccessDenied and a little logic to make the views aware if an access token involved in the request.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -42,9 +42,9 @@ class DownloadsController
 
   def handle_access_denied
     if current_user
-      error_code = '401'
-    else
       error_code = '403'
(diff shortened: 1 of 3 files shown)


app/controllers/downloads_controller.rb
```diff
@@ -38,6 +40,19 @@ def can_download?
     end
   end
 
+  def handle_access_denied
+    if current_user
+      error_code = '401'
+    else
+      error_code = '403'
```
dbrower: I think the 401/403 distinction is reversed. return 403 in then, 401 in else.

PR #290, opened 2015-11-19
Make single-use URLs valid for a time window

danhorst:
Expiring a one-time-use token after a single request is not a sufficiently sophisticated solution. For example, Chrome uses range requests to incrementally fetch PDFs. The prior implementation of this feature would deny all rage requests after the first request resulting in Chrome rendering a "empty" PDF.

Insead, “single-use” URLs will be valid for 24 hours after they have first been used to access a file.


danhorst:
@dbrower when the tokens are first created the expiry date is null. The expiry date is set a specified interval of time in the future after the first time the token is used to access a file.

A token is valid for a given SHA and noid if the expiry date is null _or_ if the the expiry date is in the future.

PR #292, opened 2015-11-20
Enhancements to access denied pages

danhorst:
Based on feedback from Mandy Havert Shari Hill Sweet:
- Give patrons with an expired access token a way of requesting renewed
  access
- Simplify messaging on the page


jeremyf:
:+1:. Waiting on travis to pass.

PR #318, opened 2016-01-21
Replace old citation class to use citeproc gem

dbrower:
In process changed how old class was collection metadata...it was
putting the advisor and contributors as the author??

I'm sure this will require more attention in the future, in particular
with the metadata passed to the citeproc gem.

[Don't merge yet]


dbrower:
@danhorst This is the branch I'm using for the citation pop-up. As I said in the stand-up, open questions are:

 * what route shall be used?
 * does the service return html or json. We decided html?

danhorst:
:+1:

PR #320, opened 2016-02-01
Show page consistency

danhorst:
This PR accomplishes two things:

1. Provides a meaningful title for the common objects viewer (which is now the only way to see ETDs)
2. Moves the page action area of the common objects viewer from a form-style line of actions at the bottom of the page to just beneth the title. This means that patrons will not have to scroll to see if/what actions are available to them on that page.

### Action area before:
move-show-page-actions-1-before

### Action area after:
move-show-page-actions-2-after

It _should_ be more visable.

### DIFF
--- a/app/views/common_objects/show.html.erb
+++ b/app/views/common_objects/show.html.erb
@@ -4,6 +4,12 @@
   <span class="human-readable-type"><%= curation_concern.human_readable_type %></span>
 <% end %>
 
+<% if can?(:edit, curation_concern) %>
+  <% content_for :page_actions do %>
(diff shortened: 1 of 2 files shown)


danhorst:
@cdeluca495 this is the opposite of what you suggested for the page actions area. I did it this way for two reasons:

1. It was easier to change because there is only one show view for the common objects viewer and there are several curation concern show views.
2. We have gotten feedback that it is difficult to find links/actions further down the page. I know “the fold is dead” but aligning page-level actions _near_ site-level actions seems like a good idea.

PR #330, opened 2016-02-19
Add administrative units to curation concern and remove organization

banurekha:


jeremyf:
We are merging this data and will be moving the AdministrativeUnits into a Hierarchy class/finder in Locabulary.

Issue #336, opened 2016-03-04
Administrative Unit wrap-up work


jeremyf:
# High Priority
- [x] Review ndlib/locabulary#22
- [x] Resolve the duplication of [Organization Helper](https://github.com/ndlib/curate_nd/blob/master/app/helpers/curate/organization_helper.rb) and [AdministrativeUnit Helper](https://github.com/ndlib/curate_nd/blob/master/app/helpers/curate/administrative_unit_helper.rb). All helper methods are mixed into the View layer; So like named methods obliterate each other. Are we using the Organization helper/information anymore?
- [x] We should be persisting to Fedora the unmodified term_label for the administrative_unit
- [x] Remove the deposit label from Locabulary; We are not using this
- [x] Remove the ["-Non-departmental" logic from AdministrativeUnits](https://github.com/ndlib/curate_nd/blob/master/app/models/administrative_units.rb#L79). Since we are persisting the label in Fedora, we need the exact label that we present in the UI and the value for the select option to be a Locabulary term_label.

# Medium Priority
- [x] `Locabulary::Item` should not have `affiliation` and `grouping` for all data types (i.e. a Copyright item should not have `grouping`).
- [x] Make a subclass of item `Locabulary::Items::AdministrativeUnit` that extends `Locabuarly::Item` and adds the `affiliation` and `grouping`
- [x] During `Locabulary.active_items_for`, leverage a lookup table to determine which class to use (i.e. `Locabuary::Items::AdministrativeUnit` when we are processing `administrative_unit` the default is `Locabuary::Item`
- [x] Move the [AdministrativeUnits](https://github.com/ndlib/curate_nd/blob/master/app/models/administrative_units.rb) logic from CurateND into Locabuary. What I'm after is repeatable delivery of the hierarchy logic; There are other hierarchies that we will need.

PR #340, opened 2016-03-08
336 administrative unit wrap up work

jeremyf:
## Removing files that are no longer used

@a4262d6d00b680ac8f95f09bfc7e4bc978a51fda

With the replacement of Organizations with AdministrativeUnits, we no
longer need the helper method nor the form partial.

## Renaming methodfor global namespace of helpers

@a46821f6f55e5427a4a0a91b6ff254f08a809814

Helper methods should be viewed as unique across all other helper
method modules. As such, I'm adding a prefix to indicate what these are
about as well as to reduce the chance of method name collision.

## Rearranging helper methods for public and private

@9ca7af359199972b7c1f974d6eb635fa3ad3c17a

Considering that only one of the methods is called in public, I want to
further isolate the exposure of helper methods.

Related to #336

## Removing uncalled AdministrativeUnits method

@2694dbd5ea130d6630f271f4f5c1e7f3270fff0f


danhorst:
:+1:

PR #342, opened 2016-03-15
Department and administrative cleanup

### LINKED ISSUE #336: Administrative Unit wrap-up work

jeremyf:
## Removing Department class as it is unused

@ab6c821f445a65377e132f14bd0b28f8d20f0268

It no longer makes sense as it is not reference throughout the entire
CurateND code base.

It should have been removed as part of the following commit:

@a4262d6d00b680ac8f95f09bfc7e4bc978a51fda

> Removing files that are no longer used
>
> With the replacement of Organizations with AdministrativeUnits, we no
> longer need the helper method nor the form partial.

## Replacing unneeded method alias

@92fbba2afb6dcebcbf053e36ee247035ce89347f

Instead of `eligible_for_selection?` which aliases `selectable?` which
aliases `selectable`, I'm removing the superfluous
`eligibile_for_selection?`

## Arranging and tightening up method names

@ad9048adad8ab1597c58ed1178422e711644a1f3

This is a very low level refactor preferring alias instead of
alias_method (as per a Rubocop suggestion). Also preferring a proper
fetch and default value instead of the inferred `nil == false`.

## Switching to hierarchy usage of Locabulary

@fb8afd68b07048eacb293cb3b1d31a4cc403b617

This does not tidy up the nasty HTML generation that is going on.
However it does normalize how we build our hierarchical Locabulary
data usage.

Closes #336

## Removing the freezing of Time via Timecop

@27a33e222f997cd59507dd3de48358bdd8c60a6d

At one point, temporal control was vital. However that does not
appear to be the case.


danhorst:
:+1:

PR #352, opened 2016-03-18
350 Add Creative Commons 4.0 Licenses

danhorst:
Additional content license options as requested by @dbrower.


jeremyf:
At a future date, we can leverage Locabulary's copyright option. But this is a good intermediary.

PR #369, opened 2016-04-15
358 include admin units in the general search for curate

banurekha:


danhorst:
I verified the desired behavior in development. I had to manually copy in the config change for my development solr:
```
cp -r solr_conf/conf/* jetty/solr/development-core/conf
```

PR #371, opened 2016-04-18
Replacing YAML config with class constant

jeremyf:
The purpose of the YAML config was to allow other applications that
mounted the Curate gem to configure their document types. Given that we
have severed the relationship with the Curate gem, we no longer need a
YAML configuration, additional initializer, and a fetch sequence.
Prefer instead the constant.


app/repository_models/document.rb
```diff
@@ -7,8 +7,13 @@ class Document < GenericWork
   self.indefinite_article = 'an'
   self.contributor_label = 'Author'
 
+  DOCUMENT_TYPES = [
+    'Book', 'Book Chapter', 'Brochure', 'Document', 'Letter', 'Manuscript', 'Newsletter', 'OpenCourseWare', 'Pamphlet',
+    'Presentation', 'Press Release', 'Report', 'Software', 'White Paper', 'Video'
+  ].freeze
+
```
dbrower: Put each element on its own line.

PR #398, opened 2016-05-03
397 Add document subtype of Catholic Document 

danhorst:
This should allow @dbrower to demarcate Catholic materials.


dbrower:
The next person to touch this list should put each item on its own line. It will make diffs nicer.

PR #405, opened 2016-05-09
404 server error when uploading new image file

danhorst:
Submitting a form for a work with more than one individual editor caused an application editor.

The `_linked_record_editors` partial was on updated improperly in 0462056c593f187fadb6a0c4a52387d04e06f1e4.


danhorst:
I verified the correct behavior on libvirt6 (https://libvirt6.library.nd.edu/concern/images/2z10wq0065h).

PR #407, opened 2016-05-11
#403 Administrative unit update to add catholic organizations

### LINKED ISSUE #403: Administrative Unit Update to Add Catholic Organizations
Working through [new protocol](https://docs.google.com/document/d/1u4lCrX6xwSORNgVWtR-0He4mX0S2xW3vB3oJDB5uJro/edit) to update Administrative Units from Administrative Unit Spreadsheet. Added new units to [spreadsheet](https://docs.google.com/spreadsheets/d/1oBW5FCTtYXsUi8roBiMRBLFY3dXamhTqy-kiG2rqu5Q/edit?ts=56ec1dda#gid=0). Filing GitHub ticket.

However, this particular update is more complicated than most because we've encountered our first top-level addition. Pat Lawton would like to be able to add things from the CRRA. Other small Catholic institutions (not "University" level like St. Mary's) have expressed interest as well. It seems most logical to create a single top-level unit for "Catholic Organizations" to group them all together for metrics and to keep them from getting scattered throughout the dropdown.

Differences of note:

1. Display in Administrative Lists is currently of the third level of faceting with the second level as header. These would have the top level as header and only be second-level as we're not going to get down to departmental stuff for any of them---most don't have departments, if they're that big, they should probably be separately classed.
2. The classification for "Catholic Organizations" is not going to be "University" because it doesn't make sense. I've suggested "OrganizationalGroup" with "Organization." If we ever wanted to add a grouping of another kind of organizations, we could do that the same way.

However these should be checked to see how much they'll affect because it's not as straightforward as simply saying "Oh Arts & Letters added a Department of Dog Photography."

danhorst:
## Claiming issue 403

@15dc9899e0c5fbd5f0c31700eebfb8081cd3a89f

relates to #403

$ ./script/start-issue 403

[skip ci]

## Accommodating more than one root Administrative Unit

This isn’t pretty but it works.

@7dba81a29900a930def623bb869399931ea3a0be

## Closing issue 403

@86299dd924a4181984bd2bf9e08f698835e7b100

Closes #403

$ ./script/close-issue 403


danhorst:
The version of locabulary will need to be incremented for the build to succeed.

PR #426, opened 2016-05-27
395 increase size of the thumbnail image on the show page

banurekha:


jeremyf:
@banurekha Does this require that we re-thumbnail all of our assets? @danhorst Thoughts?

danhorst:
I don't see an immediate need for this. The change was primarily meant to
improve support for forthcoming collections.

PR #429, opened 2016-06-02
Administrative units hierarchy facets

banurekha:


app/controllers/catalog_controller.rb
```diff
@@ -143,6 +143,19 @@ def index
     super
   end
 
+  def hierarchy_facet
+    @pagination = get_facet_pagination(params[:id], params)
+    (@response, @document_list) = get_search_results
+    respond_to do |format|
+      # Draw the facet selector for users who have javascript disabled:
+      format.html
+      format.json { render json: render_facet_list_as_json }
```
jeremyf: Is this a blacklight method?

app/controllers/catalog_controller.rb
```diff
@@ -143,6 +143,19 @@ def index
...
+
+      # Draw the partial for the "more" facet modal window:
+      format.js { render :layout => false }
```
jeremyf: What is rendered in this case? Do we use it?

app/controllers/catalog_controller.rb
```diff
@@ -143,6 +143,19 @@ def index
     super
   end
 
+  def hierarchy_facet
```
jeremyf: I'd like a corresponding HTML and JSON spec for this. Can you leverage the existing catalog controller specs?

I'm interested in the page successfully rendering when we have a work with a nested hierarchy.

app/controllers/catalog_controller.rb
banurekha: This is new action for hierarchy facets to be rendered in modal window. I have deployed this branch in libvirt8 to verify the action

app/controllers/catalog_controller.rb
banurekha: ok will add the spec

ruthtillman:
Styling/JS feedback on Admin Units cc @danhorst:

1. No help text needed for now. If people give feedback, we may add some.
2. Option to expand/collapse at Highest Level and Second level. (will this cause a problem with the non-ND two-level ones?)
3. Add "Expand All/Collapse All" toggle?
4. I can't believe I'm asking "can the bullets be prettier or...bolder" but... maybe - or > vs. normal bullets? Mostly it's that I'm not into the "circle" at College level and the 'disc' at University of Notre Dame level feels anticlimactic.

danhorst:
Mockup for point 4, improve contrast in facet hierarchy:
curatend-439-facet-hirearchy-styles
(shared with @ruthtillman in another channel)

dbrower:
@danhorst I like both the font variations in the hierarchy and also pulling the counts over to the right hand margin.

danhorst:
Hierarchical facet UI improvements:

- [x] Improve contrast of different levels in hierarchical facet modal window
- [ ] Expand/collapse behavior for leaf nodes in hierarchical facet modal window
- [x] Update facets with appropriate open/close behavior
- [ ] FIX: remove hierarchical facet link in modal window
- [x] Decorate display of selected hierarchical facets
- [ ] Add decorated hierarchical facet value to search index partial
- [ ] Update hierarchical facet value on show view to include links to searches with selected facets

PR #431, opened 2016-06-03
Updating EventMachine

danhorst:
The old version was having trouble installing on El Capitan.
Thin still works.


danhorst:
@ruthtillman this should get you back on track.

PR #438, opened 2016-06-21
Wording change for asynchronous download button

danhorst:
Suggested by @rickjohnson 

Rick was looking for a way to strongly indicate “click this button to download” while remaining concise.

Thoughts @dbrower @ruthtillman?

### DIFF
--- a/app/views/curation_concern/base/_related_files.html.erb
+++ b/app/views/curation_concern/base/_related_files.html.erb
@@ -92,7 +92,7 @@
                 target: '_blank'
               }
             ) do %>
-              <i class="icon icon-white icon-time"></i> Request Retrieval
+              <i class="icon icon-white icon-time"></i> Initiate Download
             <% end %>
           <% else %>
             <%= link_to(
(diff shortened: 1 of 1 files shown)


dbrower:
"Initiate Download" sounds like it will start a download immediately, not after an arbitrary amount of time.

"Get from Tape"? "Retrieve from Tape"? "Initiate Retrieval"? "Retrieve"? "Wait for Download"?

ruthtillman:
Yeah this is my concern as well. "Start Retrieval for Download"? So long.
"Retrieve for Download"? Still too long?

rickjohnson:
'Retrieve for download' is definitely closer.  'Fetch for Download'? (three fewer characters lol).  Some kind of message above the buttons in the layout might also close the gap like "This file is not immediately accessible.  Click 'Retrieve for Download' to make the file available for download".

rickjohnson:
Just 'Retrieve File' might be an alternative as well (possibly with the help message I proposed in my last comment)

jeremyf:
Per conversations with Kate R, a previous editor and UX person, "Begin Download" with an indicator that it could take a bit (we have a tooltip and clicking the link informs the user of the delay for download).

jeremyf:
@dbrower @ruthtillman Anyone care to push the tempting green "Merge pull request" button?

PR #440, opened 2016-06-22
420 collections modeling

jeremyf:
## Adjust Collection Behavior to include new set of precidate as per new specification

@3b7648832af6397cb669e2ec437db45d6c883254

As part of Project_17, this commit will include new set of attributes in collection

Add/Adjust new attributes to descMetadata
Update form to include those attributes
Show them in collection show page

## Adding spec for collection model

@7170593b24b6cab252b47f407854d4a0fc516758

## Removing additional descMetadata added to collection form and adding spec for collection

@841b0f61cd075b25405ffe4e88ac990fad21371d

## Remove edit and delete collection feature from UI

@35a401dc3396ab8518d8fb17b072bc9b2d2c1cb0

## Adding curate's collection specs

@d08f1559eebc931770da0f5195583742acf3bc89

These needed to be ported over from Curate's test suite before we
begin gutting the collections.

## Renaming spec file to reflect convention

@5bc75e9e1a97cc2ce1c3809e962ee0bff79b97e7


app/models/ability.rb
```diff
@@ -22,4 +23,158 @@ def custom_permissions
...
+      p.pid == current_user.repository_id
+    end
+
+    can [:show, :read, :update, :destroy], [Curate.configuration.curation_concerns] do |w|
+      u = ::User.find_by_user_key(w.owner)
+      u && u.can_receive_deposits_from.include?(current_user)
+    end
+  end
+
+  def licensing_permissions(licensing_permission = ContentLicense.new(current_user))
```
dbrower: Is the `current_user` at def time the same as the `current_user` when the function is called?

jeremyf: Yes.

PR #449, opened 2016-06-28
410 do not pull title from pdf metadata

### LINKED ISSUE #410: Do not pull title from PDF metadata
At some step in the characterize process, the worker pulls the title from the PDF metadata and sets the file's title to that. The title is then displayed before the file on the Work show page. The problem is that this title is usually gibberish.

The issue is that the file name should be displayed on the Work show page. I suppose it is enough to change that page to display the file name instead of the title.

I don't have any nice pictures.

jeremyf:
## Claiming issue 410

@dfee9a97c94015008effb4baa6a820745530c31e

relates to #410

$ ./script/start-issue 410

[skip ci]

## Replacing the usage of Title with Filename

@d2ca72ad41999a869a51341535c3cce3f1e44393

In reviewing the behavior with Don, we don't want to show the extracted
title from fits. But instead show the filename.

## Closing issue 410

@775f87d12e5c6301ba5ce3b95f4841a85bdbeaec

Closes #410

$ ./script/close-issue 410


dbrower:
Does the partial for the file list on a work's show page need to be changed as well?

jeremyf:
@dbrower No. The rendering partial is using https://github.com/ndlib/curate_nd/blob/e17c0fe965ffb17ee0f60350aa9c5cfee4a51d53/app/views/curation_concern/base/_related_files.html.erb#L28

Which defers to the generic file's `#to_s` method.

PR #450, opened 2016-06-28
411 only enforce future embargo dates for new records

### LINKED ISSUE #411: Only enforce future embargo dates for new records
If you create a record with an embargo date in the future, wait until the embargo date has passed and then try to update the record you encounter the error “Must be a future date” for the embargo date.

You should be able to leave the original access rights metadata intact even though the record will be treated as Open Access now that the embargo date has passed.

Moreover, even if you select “Open Access” instead you still receive the same error about the need for an embargo date to be in the future.

Example record: https://curate.nd.edu/show/bk128912p66

laritakr:
## Closing issue 411

@3df4ade03d40ce24cf62e723924ecb6429f8b1f2

Closes #411

$ ./script/close-issue 411


laritakr:
Note: The solution allows any valid date as embargo date, and does not differentiate between new and changes as implied in the issue title.

dbrower:
Should the `future_date_validator` code be removed? Is it still being used?

laritakr:
No it isn't, and I asked that question as well. Waiting for an answer on it.

PR #469, opened 2016-07-12
468 display book authors/LCSH in search result

ruthtillman:
Improved search result display for Document type Book objects. Went through logic with @danhorst before deploying but wouldn't hurt to have another pair of eyes on it.


dbrower:
My only comment is whether the check for the book type is needed? Maybe just remove lines 16 and 26. Display authors, if present. Display creator if present. If both are present, then the "Author(s)" label will show up twice. Presumably that would be nice to know since it is probably bad metadata? And no metadata is suppressed.

ruthtillman:
I really don't want "Author(s)" showing up twice if that ever happens. It would look really gross. :-[

dbrower:
I agree, it would look gross. And it would indicate that something is wrong.

@danhorst could values from two different fields be put in the same `<dd>` tag?

ruthtillman:
@danhorst @dbrower again, that requires the double test to see if OR exist... if we can write that OR test I don't have a problem with it. But it was already discussed.

dbrower:
@ruthtillman is the OR test "is there an author OR creator (OR both)"? I missed the previous discussion, so I will defer to that.

ruthtillman:
@dbrower Essentially that, yes. Apparently the method of querying the solr doc doesn't allow for complex boolean stuff.

PR #473, opened 2016-07-14
always refresh disadis executable when puppet is run

msuhovec:
#472


puppet/modules/lib_curate/manifests/disadis.pp
```diff
@@ -16,9 +16,6 @@
 		environment => ["GOPATH=${disadis_root}","GO15VENDOREXPERIMENT=1"],
 		path => "/usr/bin",
 		command => "go get github.com/ndlib/disadis",
```
dbrower: Might need to use `go get -u`.

PR #483, opened 2016-07-22
Shifting LibraryCollection representative object

jeremyf:
> As a Librarian, I want a standard way to assign representative images
> to Library collections, so that I can use the batch ingester in a
> predictable way.

We want to normalize where the representative image is located (i.e. in
the properties datastream). This aligns LibaryCollections with
Documents, Images, etc. It also allows for the eventual migration to a
more appropriate data model.

### DIFF
--- a/app/repository_models/library_collection.rb
+++ b/app/repository_models/library_collection.rb
@@ -11,7 +11,7 @@ class LibraryCollection < ActiveFedora::Base
   include CurationConcern::HumanReadableType
 
   has_metadata "descMetadata", type: LibraryCollectionRdfDatastream
-  has_metadata "properties", type: Hydra::Datastream::Properties
+  has_metadata "properties", type: Curate::PropertiesDatastream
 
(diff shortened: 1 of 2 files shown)


dbrower:
I have questions on the implications of this...namely, can a collection have a generic file directly attached to it? for the thumbnail, ya know?

jeremyf:
@dbrower That appears to be how @ruthtillman set up the example collection on Libvirt9

https://libvirt9.library.nd.edu/show/dev00144159

ruthtillman:
I think the answer is...yes? That's how we set it up, submit the collection with a representative file that's pretty much always an image because there's no real mechanism for accessing it.

dbrower:
This is fine. It is different to how the collections were stored, but i don't see any problems with it. e.g. this fits the PCDM model more or less.

ruthtillman:
Yeah, we noticed the differences with how collections had handled their objects but I think having everything in alignment right now is good and would help us transition more smoothly in future.

PR #487, opened 2016-07-27
Remove index_type for field dc:source

laritakr:
As Don requested, I took out index.type :text because this is a URL.


dbrower:
I would merge this, but I'm unsure of the implications with SOLR. @jeremyf what are the ramifications of this change? will things need to be reindexed in solr?

jeremyf:
@dbrower I believe you are correct that EADs will need to be reindexed. I believe this means we'll need a task as part of the post deploy. I'll make a note of what needs doing on the JIRA issue.

dbrower:
@jeremyf I think the patents are also using `dc:source`. There may be a few others items as well.

ruthtillman:
Patents do. I don't know if anything else necessarily does, but it's possible they may.

Issue #470, opened 2016-07-13
 Ingested Curate Items modified by the GUI needed to be detected and reingested via the Batch Ingest system


msuhovec:
See JIRA DLTP-479  https://jira.library.nd.edu/browse/DLTP-479   or
Project 30: Bendo Write Integration https://docs.google.com/document/d/1V6MHVjpcJ4G_QGknqMTBndbuCAtagix1wRPHuxpYcfs/edit#

for more information

PR #488, opened 2016-07-29
Issue 470

msuhovec:
#470


app/services/batch_ingest_tools.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    make_job_id
+    create_batch_job
+    add_job_file('fedora-pids', pid_list)
+    submit_batch_job
+
+  end
+
+  def self.make_job_id
+    @job_id = Time.now.strftime 'reingest%Y%b%e%H%M'
```
dbrower: add a random number to the end just to make sure we don't end up with duplicates. so, e.g. `reingest2016jul291532-12345`

lib/tasks/bendo.rake
```diff
@@ -0,0 +1,25 @@
+namespace :bendo do
+
+  timestamp_file = '/home/app/curatend/shared/system/lastrun'.freeze
```
dbrower: maybe set the timestamp file using an environment variable? then one can also run this in testing and on a laptop. The cron job can then set where the file is `env TIMESTAMPFILE=/a/b/c rake bendo:sync_with_fedora`

msuhovec: I created a JIRA ticket ,https://jira.library.nd.edu/browse/DLTP-520, to
track this. I may need to add some info to secrets

app/services/batch_ingest_tools.rb
msuhovec: I added %s - now it's microsecond granular

job_id = Time.now.strftime 'reingest%Y%b%e%H%M%s'
=> "reingest2016Jul2916401469824806"

app/services/batch_ingest_tools.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  end
+
+  def self.make_job_id
+    @job_id = Time.now.strftime 'reingest%Y%b%e%H%M%s'
+  end
+
+  # does PUT /jobs/:jobid to create data dir on batch ingest side
+  def self.create_batch_job
+    url = 'http://localhost:15000/jobs/' + @job_id
```
jeremyf: Is this the long-term URL for the create batch job?

msuhovec: yes- it will always running on same box as batch ingest- the cron job that
will invoke this has not yet been written

dbrower:
@jeremyf LGTM

Issue #447, opened 2016-06-28
Sort departments on departmental landing page according to order in locabulary


danhorst:
Added to JIRA - https://jira.library.nd.edu/browse/DLTP-512

The `/departments` page is built using the administrative unit facet values. Instead of listing the terms in alphabetical order these terms should be ordered according to their sequence in [locabulary](https://github.com/ndlib/locabulary). This will require enhancing or replacing the behavior in [`Blacklight::HierarchyHelper`](https://github.com/ndlib/blacklight-hierarchy/blob/master/app/helpers/blacklight/hierarchy_helper.rb).

- [x] Sort administrative units in “locabulary order“

Once the administrative units are sorted in “locabulary order“ `/departments` page should be limited to _only_ units in the University of Notre Dame. This should be accomplished with CSS scoped to the full page view. That way all the values can still be present on the facet modal view.

- [x] Hide all non-ND units on the `/departments` page

## Before
![curatend-447-sort-departmental-landing-page-before](https://cloud.githubusercontent.com/assets/2133/17176909/d80ad7be-53dc-11e6-91d3-e1fc64a1217e.png)

## After
![curatend-447-sort-departmental-landing-page-after](https://cloud.githubusercontent.com/assets/2133/17176913/de718e86-53dc-11e6-8417-75b15c9c4f03.png)

ruthtillman:
And once sorting can be done, this page can be sure to only display things under University of Notre Dame, yes? Probably doesn't actually need University of Notre Dame header since page name "Departments and Units" does the trick.

danhorst:
This might be possible. It would be easier to keep the content on the page
and in the facet selection window the same. That way I don't have to
conditionally filter out things from CRRA etc.

ruthtillman:
I have a pretty strong feeling people would rather that be suppressed. Thinking logistically, if it's still coming from a place of top-level ol with nesting, Notre Dame now sorted to be on top, etc., would there be a way to use CSS to suppress display of everything but first child, aka ND? That's not a thing I really want to be saying and it may be fever-logic.

danhorst:
That makes sense. If hiding content with CSS rather than excluding it from the markup entirely is acceptable it simplifies the implementation dramatically. `:first-child` FTW.

danhorst:
Closed via PR 490

PR #490, opened 2016-08-08
Implementing sorted departmental landing page

jeremyf:
This can be verified on Libvirt9 (https://libvirt9.library.nd.edu/departments) as of August 8, 2016.

## Updating Locabulary to include Hierarchy processing

@99e7680fcab1c8c6abe06b08520d0ebe3ffd1715

## Implementing sorted departmental landing page

@9df90af21c4a9e45958d4a916f9c1cf123d3f36e

By moving the logic into the presenter, we can skip the custom
rendering of Blacklight Hierarchy and instead encode that information
in a class. This allows for more isolated testing as well as, what I
hope to be, a more clear description of what is going on.

DLTP-512

## Customizing interception of department facet URL

@132864484488a3cb38e7b13718c63a638e84f977

There are two places that should change:

1) The Facets on the left
2) The /departments landing page

With this change, we address item 1. The previous commit addressed
item 2.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -35,12 +35,12 @@ GIT
 
 GIT
   remote: git://github.com/ndlib/locabulary.git
-  revision: 999fa84b5bb3f973d71ac8a6a02e34bb2499ed36
+  revision: aec912e2d617788d061df9511d5598ef58d020fb
   branch: master
   specs:
-    locabulary (0.4.0)
+    locabulary (0.6.0)
+      activesupport (~> 4.0)
       dry-configurable
-      hanami-utils
       json (~> 1.8)
 
 GIT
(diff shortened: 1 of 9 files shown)


danhorst:
While this creates the /departments page correctly it does not update the facet partial. This is probably OK.

We need a way to hide non-ND departments and units on the listing page. I will do this via CSS.

jeremyf:
@danhorst Which facet partial are you referring to? I can make updates there.

What are the non-ND departments that should be omitted? Catholic Organizations?

danhorst:
I was referring to how facets are rendered in the catalog view:
curatend-pull-490-search-view
curatend-pull-490-selected-facet

Facet configuration is in the `catalog_controller`

ruthtillman:
Clarity note: In search results sidebar faceting, the removal of Catholic Orgs, etc., is unnecessary. However for the Departmental Landing Page, we want them not displayed.

jeremyf:
@ruthtillman I think Dan is working on that in another step.

jeremyf:
@ruthtillman The remaining work for not showing Catholic Organizations on the /departments page is account for in [DLTP-529](https://jira.library.nd.edu/browse/DLTP-529)

ruthtillman:
Awesomesauce. I just wanted to make sure that difference was clear so too much work didn't end up accidentally getting done on it.

danhorst:
This addresses #447

Issue #491, opened 2016-08-10
Change BatchIngestTools to queue start-reingest task


msuhovec:
A single batch ingest task, start-reingest, will queue all subtending reingest activities, and BatchIngestTools should create a JOB file in the jobqueue directory containing this task in the Todo list.

PR #492, opened 2016-08-10
change BatchIngestTools to queue start-reingest task

msuhovec:
#491

### DIFF
--- a/app/services/batch_ingest_tools.rb
+++ b/app/services/batch_ingest_tools.rb
@@ -6,16 +6,13 @@ require 'uri'
 # Use Batch Ingest HTTP Interface to submit modified fedora objects
 # that contain bendo items to batch ingest system, for possible reingest
 class BatchIngestTools
-
   def self.submit_pidlist(pid_list)
-
     make_job_id
     create_batch_job
     add_job_file('fedora-pids', pid_list, false)
     task_list = make_task_list
     add_job_file('JOB', task_list, true)
     submit_batch_job
-
   end
 
   def self.make_job_id
@@ -26,8 +23,8 @@ class BatchIngestTools
   def self.make_task_list
     list = {}
     list['Todo'] = []
-    list['Todo'].push( 'start-reingest' )
-    return list
+    list['Todo'].push('start-reingest')
+    list
   end
 
   # does PUT /jobs/:jobid to create data dir on batch ingest side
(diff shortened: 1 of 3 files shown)


app/services/batch_ingest_tools.rb
```diff
@@ -6,20 +6,27 @@
...
 # that contain bendo items to batch ingest system, for possible reingest
 class BatchIngestTools
-
   def self.submit_pidlist(pid_list)
-
     make_job_id
     create_batch_job
-    add_job_file('fedora-pids', pid_list)
+    add_job_file('fedora-pids', pid_list, false)
+    task_list = make_task_list
```
dbrower: Why not do the following?

```
task_list = {'Todo' => ['start-reingest']}
```

msuhovec: I knew you'd want that.

app/services/batch_ingest_tools.rb
```diff
@@ -30,23 +37,27 @@ def self.create_batch_job
   end
 
   # does PUT /jobs/:jobid to create data dir on batch ingest side
-  def self.add_job_file(name, content)
-    url = 'http://localhost:15000/jobs/' + @job_id + "/files/" + name
+  def self.add_job_file(name, content, want_json)
+    url = 'http://localhost:15000/jobs/' + @job_id + '/files/' + name
```
msuhovec: That was my intent- it seemed more palatable than passing it as a parameter
or using a get_job_id method in each of the methods that call it. I would
guess the Ruby
community at large would scoff at me using class methods at all.

app/services/solr_tools.rb
```diff
@@ -51,7 +54,7 @@ def self.solr_url
     @solr_url[Rails.env]['url']
   end
 
-  #build query 
-  def self.query_params(fromtime, totime)
+  # build query
+  def self.query_params(_fromtime, _totime)
   end
```
dbrower: What is this function doing? why not remove it?

dbrower:
For the timestamp file, consider that both the rake task and the ingest tool have to understand how the timestamp file works: The modification time of the file indicates the last time the harvest was run. Wouldn't it be simpler if only the rake task knew about the timestamp file and its mtime semantics, and it passed an optional date parameter to the ingest tool, the parameter indicating the time to start looking for changes.

msuhovec:
I thought about this. I guess I like my rake tasks to be big, dumb
supervisor, and encapsulate logic in the class they called. The rake knows
where it lives, and to update it. That is all.

 From the maintenance standpoint you can see on one line in
BatchIngestTools that I use the modification time, in UTC, converted to
iso8601 format for SOLR comparion  purposes

dbrower:
But then the rake task touches the file! Why doesn't the ingest service touch the file when it is finished?

Maybe it does come down to a difference in philosophy. I noticed this inconsistency. And then to resolve it I thought "the only aspect of the file that is needed is its mtime". And so, rather than pass the file around, just passing a time would mean I can remove any knowledge of timestamp files from the ingest code. I'm okay with rake tasks being complicated.

msuhovec:
I'm OK with moving the touch out of the rake down into solr_tools.

dbrower:
no need. I was only pointing it out to illustrate that the rake task is not that dumb.

app/services/solr_tools.rb
msuhovec: deleted.

PR #513, opened 2016-08-22
Change start-issue and close-issue to support JIRA issues

dbrower:
Support for github issues is still there, but if a jira issue
is detected, it will try to update the jira ticket and will even resolve
it. The details of the workflow are easy to change.

All the interaction with jira are done using a command line tool `go-jira`.
It can be installed on macs using `brew install ndlib/dlt/go-jira`.
The tool was designed to be used by hand, there is some weirdness in
calling it from a script, mainly because one needs to make sure a current
password cookie is saved.

DLTP-563

### DIFF
--- a/script/close-issue
+++ b/script/close-issue
@@ -6,12 +6,10 @@
 #
 #*******************************************************************************
 
-CONFIG_KEYS = [:REMOTE, :REPOSITORY_PATH, :STARTED_ISSUES_FILE].freeze
+CONFIG_KEYS = [:REPOSITORY_PATH, :STARTED_ISSUES_FILE].freeze
 
 REPOSITORY_PATH = ENV.fetch('REPOSITORY_PATH') { File.expand_path(File.join(File.dirname(__FILE__), '../')) }
-REMOTE = ENV.fetch('REMOTE', 'origin')
 
-# TODO: Retrieve the dasherized issue from Github's API
 STARTED_ISSUES_FILE = ENV.fetch('STARTED_ISSUES_FILE', '.started-issues')
 
 #*******************************************************************************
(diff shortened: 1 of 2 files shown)


dbrower:
There seems to be some room to fiddle with the exact statuses the JIRA tickets are transitioned to. In the PR as written, using `start-issue` will move a JIRA issue to _In Progress_ and then using `close-issue` will move it to _Deployed_. Should these be changed? Maybe use _QA_? I don't know. @danhorst @jeremyf feedback welcome.

jeremyf:
@dbrower Where is the source code for go-jira? It looks like something you wrote.

dbrower:
https://github.com/Netflix-Skunkworks/go-jira

I did make a homebrew formula for it.

script/close-issue
```diff
@@ -123,3 +128,8 @@ begin
 ensure
   File.unlink(path_to_commit_message) rescue true
 end
+
+# update jira and mark this issue as finished
+if issue_source == :jira
+  `jira transition deployed #{ISSUE_NUMBER} --noedit`
```
jeremyf: Should we be capturing the `$?` from this command? If so prefer `system("jira transition deployed #{ISSUE_NUMBER} --noedit")` and check for true/false.

script/start-issue
```diff
@@ -172,3 +197,9 @@ begin
 ensure
   File.unlink(path_to_commit_message) rescue true
 end
+
+# If the commit was successful, assign the issue to us and move into "In Progress"
+if issue_source == :jira
+  `jira take #{ISSUE_NUMBER}`
```
jeremyf: Should we be capturing the `$?` from this command?

dbrower: np. What should it do if the command fails?

jeremyf: Put something to `$stderr` (I believe `command` does not swallow stdout nor stderr) saying the command failed. Then `exit!(<next-integer>)` (which I believe would be 5).

dbrower: I was only wondering what the point is, since the script doesn't really have a fall back plan. And if the command doesn't exist the backtick will raise an error.

I'll change these two. Should I change all the other backticks to check for errors?

dbrower: Ok. it seems using `system` is better here since it hooks the subcommand up to the main process's `stdin`, `stdout`, and `stderr`.

This might solve the password problem...but I think I will still submit a patch upstream to display the password prompt on `stderr`.

jeremyf:
@dbrower If you want to change other backticks, that's fine. But its beyond the scope for this commit.

PR #522, opened 2016-08-31
Dltp 536 as the curate nd application i need to model the osf archive package so that curate nd can make some sense of the project registration 

laritakr:
## Claiming issue DLTP-536

@2078b31c87ac8a73b84bf35c7e612acdda1568e8

$ ./script/start-issue DLTP-536

[skip ci]

## Model OSF Archive package

@d47012aedda5ec686991e4eced8990ab9f5fb9f0

## Merge branch 'master' into DLTP-536-As-the-CurateND-application-I-need-to-model-the-OSF-archive-package-so-that-CurateND-can-make-some-sense-of-the-project-registration-

@84e5e92bb1779f93aebaf87fecabd97ed1efabe5

## Cleanup config related to DLTP-580

@3ab17dd35c2a1de84da0ab955c4b31242b094fd4

Staging files were changed, but corresponding CurateND files needed some
related cleanup for consistency purposes.


app/repository_models/osf_archive.rb
```diff
@@ -0,0 +1,83 @@
...
+  self.human_readable_short_description = "Change me."
+
+  def set_default_values
+    self.type = 'OSF Archive'
+  end
+
+  def set_initial_values
+    self.date_modified = Date.today
+    self.date_archived = Date.today
+  end
```
jeremyf: I don't think this method should be exposed to other objects. Please add `private :set_initial_values`

app/repository_models/osf_archive.rb
```diff
@@ -0,0 +1,83 @@
...
+  include CurationConcern::RemotelyIdentifiedByDoi::Attributes
+
+  after_initialize :set_default_values
+
+  before_validation :set_initial_values, on: :create
+
+  class_attribute :human_readable_short_description
+  self.human_readable_short_description = "Change me."
+
+  def set_default_values
```
jeremyf: I don't think this method should be exposed to other objects. Please add `private :set_default_values `

PR #523, opened 2016-09-01
Ingest OSF Project UI

jgondron:
This adds the UI needed to create an ingest job that will (eventually) import data from an OSF project.

Summary of changes:
- Created a new model to encapsulate the properties of the request. The model has no associated database entity as this will just be passed through to the batch ingest system.
- Created a controller to handle requests for ingesting new projects and showing existing ingest jobs. 
- The new form uses existing partials such as the administrative unit form. I modified this to allow rendering as a required input, so this change will affect other views and should be QA'd with this in mind.
- It does not actually add the job to the batch ingest system. For now, the creation and querying of OSF jobs is mocked out in the IngestOSFTools service class. This part of the code path may be rewritten or moved to another service class as part of DLTP-584.

### DIFF
--- a/app/controllers/admin/ingest_osf_archives_controller.rb
+++ b/app/controllers/admin/ingest_osf_archives_controller.rb
@@ -1,0 +1,30 @@
+class Admin::IngestOsfArchivesController < ApplicationController
+  with_themed_layout('1_column')
+
+  # GET /admin/ingest_osf_archives/new
+  def new
+    @archive = Admin::IngestOSFArchive.new
+  end
+
+  # POST /admin/ingest_osf_archives
+  def create
+    @archive = Admin::IngestOSFArchive.new(admin_ingest_osf_archives_params)
+
+    if IngestOSFTools.create_osf_job(@archive)
+      redirect_to admin_ingest_osf_archives_path, notice: 'Archive ingest was successfully created.'
+    else
+      render action: 'new'
+    end
+  end
+
+  # GET /admin/ingest_osf_archives
+  def index
+    @archives = IngestOSFTools.get_osf_jobs
+  end
+
+  private
+
+  def admin_ingest_osf_archives_params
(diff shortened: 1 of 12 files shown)


app/controllers/admin/ingest_osf_archives_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  helper_method :affiliations
+
+  def curation_concern
+    GenericWork.new
+  end
+  helper_method :curation_concern
+
+  def admin_ingest_osf_archive_params
+    cleaned_params = params.require(:admin_ingest_osf_archive).permit(:project_identifier, :affiliation, :owner, administrative_unit: [])
+    match = /^https?:\/\/[^\/]*\/([^\/]*).*$/.match cleaned_params[:project_identifier]
```
jeremyf: Is there a reason for doing this in the controller? Can this be done as part of the setter method for the `Admin:: IngestOSFArchive`?

jgondron: I would rather the model be very lightweight, just an encapsulation of the data needed. This would require the model doing some things to preprocess the data in the setter, which I'd like to avoid. If you're uncomfortable with doing it here, I would say we create a service class that performs a CreateIngestOSFArchive type function that expects either a url or id for that param, and change the controller to use that instead of directly instantiating the model.

jeremyf:
@jgondron I agree with your thought process. How about creating a custom build method on `Admin::IngestOSFArchive` that performs the sanitization.

```ruby
class Admin::IngestOSFArchive
  def self.build_from_user_input(attributes)
    # perform santization/coercion
    new(coerced_attributes)
  end
end
```

This allows for testing of your logic in a model instead of in the controller.

jgondron:
That'll work. We generally try to keep all of these kinds of things outside of the model, but that's a happy medium if you guys don't follow that strictly.

jeremyf:
@jgondron I like class methods for those purposes. Especially when dealing with a simple data structure.

PR #528, opened 2016-09-07
Edit OSF Archive

jgondron:
This adds the ability to edit OSF Archive metadata.

Changes:
-Added an actor to handle modifying an OSF archive
-Added additional/required view partials based on the fields that are currently in the show view for an OSF archive


app/services/curation_concern/osf_archive_actor.rb
```diff
@@ -0,0 +1,4 @@
+module CurationConcern
+  class OsfArchiveActor < GenericWorkActor
```
jeremyf: Can you add the `remove_method :create` and add a comment about this is not how its done.

PR #541, opened 2016-09-15
DLTP-598 As a user, I would like to acknowledge I'm a postdoc in my Affiliation

ruthtillman:
Requested by Tracy. I think I've found/fixed the place where it's set up. I used grep to find the places where it's spelled out in the settings. But if I missed something, let me know.


dbrower:
Did someone mention another "affiliation" besides postdoc in that meeting? I can't remember.

ruthtillman:
I don't believe so, no. It came up when we were talking about the "Academic Rank" facet rename...

PR #550, opened 2016-09-30
Dltp 624 osf archive missing title on se

laritakr:
## Claiming issue DLTP-624

@7e0af21d0f18077b9b71b7d3a6dc5707cc14974b

$ ./script/start-issue DLTP-624

[skip ci]

## Clean up datastream fields & usage

@ce4f71cf3f8dcce56dfd26f63603d6c5b43a424a

Clear up the definition and use of the datastream fields. Change
some to not be required, set defaults, and clean up related show and
edit pages. Correct route mapping so OSF Archives use show page rather
than curation concern page.

## Fix spec factory

@4bdb6a7556f4b641369acaf2de3b8ce400f23fef

Add and remove fields corresponding to datastream changes.

## Add source to FactoryGirl

@e4d0fdc1129736e3d8c2ccbc4ad60eeabc365143

Removed the "allow blank" option so needed to initialize it.

## Closing issue DLTP-624

@38b6ded6218b728e8325bb2595dddd9f9447673a

$ ./script/close-issue DLTP-624


app/repository_models/osf_archive.rb
```diff
@@ -31,11 +31,6 @@ def set_initial_values
 
   has_metadata 'descMetadata', type: OsfArchiveDatastream
 
-  attribute :administrative_unit,
-    datastream: :descMetadata, multiple: true,
-    label: "Departments and Units",
-    hint: "Departments and Units that creator belong to."
```
msuhovec: "Departments and Units that creator belong to."  belong -> belongs ?

laritakr: I just copied it. All of the models have that same wording.

Issue #428, opened 2016-05-31
Push CSV to Share


jeremyf:
See for further details on task (https://docs.google.com/document/d/1kD4hegjUyFcB6sqFzWOY_6hpRn3rkMi29IoXXWkRkxk/edit#)

Related to #427

For each CSV row:
- [ ] Expand into JSON format that is expected by  [Share API](https://osf.io/api/v1/share/data/help/#!/SHARE/post_share_data)
- [ ] Post the JSON format to Share API
- [ ] Capture any non 2XX HTTP requests

For the job:
- [ ] If there are any non-2XX HTTP requests, Notify ERRBit of the exception

dbrower:
Again I see this as a command line program. If you see it as needing to be more complicated (eg errbit), then ok..

Most of the difficulty here will be getting an endpoint set up with the COS people.

dbrower:
I started the branch https://github.com/ndlib/curate_nd/tree/share-push-gateway and have added a few command line utilities to generate a CSV file for SHARE.

This is blocked on the SHARE people changing their API and not giving me access to the old one.

Issue #427, opened 2016-05-31
Generate CSV of changed items


jeremyf:
Generate a CSV file of items that have changed or were modified in a given time period. The field list is defined in [Support Harvesting Content to SHARE - Project 24](https://docs.google.com/document/d/1kD4hegjUyFcB6sqFzWOY_6hpRn3rkMi29IoXXWkRkxk/edit#)
- [ ] Query SOLR for date range
- [ ] Generate CSV file and write to the file system
  - [ ] Note the column names should be the same as the SHARE name
  - [ ] The file name should be `curate-to-share-from-yyyymmdd-to-yyyymmdd.csv`
- [ ] Create an asynchronous job that:
  - [ ] Run with a `:from_date`, `:to_date`, and `:recipient_emails` parameters
  - [ ] Generates a CSV
  - [ ] Sends the CSV as an attachment to the provided email address

Assumption: The first request won't be via a web request; Instead we will perform the request by hand (as it could be a rather lengthy query).

dbrower:
Is a background job needed? The make-csv-from-solr is pretty close to the level needed. And since this would be used in production only a few times, it is okay for it to require, say, me to make the file for Reid.

dbrower:
I started the branch https://github.com/ndlib/curate_nd/tree/share-push-gateway and have added a few command line utilities to generate a CSV file for SHARE.

PR #564, opened 2016-10-24
Share push gateway

dbrower:
The two scripts to at least push data to SHARE.

addresses DLTP-411
related to #428, #427

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ group :default do
   gem 'rubydora', '~> 1.7.4'
   gem 'sanitize', '~> 3.0.2'
   gem 'select2-rails'
+  gem 'share_notify', '~> 0.2'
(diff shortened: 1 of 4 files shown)


script/make-csv-of-solr.sh
```diff
@@ -72,15 +73,15 @@ first=1
 for f in $(ls results/*.csv); do
     # only copy the labels on the first file
     if [[ $first == 1 ]]; then
-        cat $f > $TARGET_ALL
+        cat $f > "$TARGET_ALL"
         first=0
```
msuhovec: I guess the spot where you increment this is somewhere in this.

dbrower: It looks like `first` is set to 1 and then changed to 0.

PR #566, opened 2016-10-26
Patching active_fedora to handle date parse error

banurekha:


lib/active_fedora/rdf_node/term_proxy.rb
```diff
@@ -0,0 +1,138 @@
...
+
+        parent.query(subject, options.predicate).each do |solution|
+          v = solution.value
+          v = v.to_s if v.is_a? RDF::Literal
+          if options.type == :date
+            begin
+              v = Date.parse(v)
+            rescue ArgumentError
+              # handle invalid date
+              ActiveFedora::Base.logger.error "Invalid Date: " + v
```
jeremyf: I believe for this review, we were hoping @dbrower or @ruthtillman could give use guidance on what we should do with an ArgumentError. The particular case is that we have an invalidate date (a triple with a "" for the date). 

* Do we want to log this behavior (to a log that is rarely reviewed) but proceed with rendering (Status 200)?
* Throw an exception and render an error page (Status 500)? (Current behavior before this pull request)
* Send a message to Airbrake capturing the exception, but proceed with rendering (Status 200)?

With that decision, I can move forward with the review.

ruthtillman: Thanks @jeremyf. After talking to Don, we would like to have it logged for eventual review but not for immediate review. I think then it's on us to review the log as a part of our metadata cleanup plan.

jeremyf: @ruthtillman With log rotation, we may only have 30 days of logs, which means we may not have hits on objects that have invalid dates. I'm okay proceeding with the current solution, but want you to make as informed a decision as possible.

jeremyf: @dbrower can you review the above?

dbrower: For sure the request should succeed when bad metadata is present. Ideally the bad data is somehow passed on in its unprocessed form, like as a string.

I don't care too much about logging. Maybe the errbit logging is better since then we will know there are bad values. But I don't see fixing them individually to be a high priority. So the same bad data may cause a lot of errbit messages. We can adjust if it is too much or not useful.

dbrower: So I basically echo what Ruth says.

PR #568, opened 2016-11-03
Dltp 662 add json ld rendering option fo

banurekha:
## Claiming issue DLTP-662

@a2302d229e63896d2a92eba1c2f4c04d9871013f

$ ./script/start-issue DLTP-662

[skip ci]

## Add json format for an item

@cc02796666a1a9046221bc548ae9d55678bca445

This change invloves creating a mapper class to map each attributes in datastreams to a key.
Also convert each datastream to given json format.

## Reworking on repository model datastreams mapper to work with all models

@65e7685c84bdb83b74a8ee010090447209ea62dd

This change is expected to work with all our repository model and all know datastreams. Also handle single object, and as a “@graph”, in the case where the descriptive metadata uses blank nodes.  Also using RDF and JSON_LD gems instead of building JSON manually. Adding respective spec for those

## Adding as_json_ld method to Article to get datstreams as JSON

@7ea88457d61fc0113780f40aaa6dec5f46cc4b8b

## Changes related to review comments

@ad487b6f57a663d978bfc8db21c1829cdcdb90c6

## Closing issue DLTP-662

@cfba663aa6b6f394e73b268a96dfbda94892afe0

$ ./script/close-issue DLTP-662


app/mappers/datastream_json_mapper.rb
```diff
@@ -0,0 +1,271 @@
+require 'json'
+require 'rexml/document'
+require 'rdf/ntriples'
+require 'rdf/rdfxml'
+  # Mapper to map datastreams to JSON format
+  class DatastreamJsonMapper
+    CONTEXT = {
```
banurekha: ROF has the context split into two (RDFContext and RelsExtRefContext) and curate conversion need to have them in one place. Also this will introduce ROF dependency, curate may not be in sync with ROF gem dependency.

app/mappers/datastream_json_mapper.rb
```diff
@@ -0,0 +1,271 @@
...
+      new(curation_concern, **keywords).call
+    end
+
+    def initialize(curation_concern,attribute_map: default_attribute_map)
+      self.curation_concern = curation_concern
+      self.attribute_map = attribute_map
+      self.fedora_info = {}
+    end
+
+    def call
```
banurekha: Make sense! Thought about it as well and will do

app/mappers/datastream_json_mapper.rb
```diff
@@ -0,0 +1,271 @@
...
+      process_datastreams
+      #merge_strip_blank values
+      return format_output
+    end
+
+    def process_datastreams
+      curation_concern.datastreams.each do |dsname, ds|
+        next if dsname == 'DC'
+        method_key = dsname.sub('-', '').to_sym
+        if self.respond_to?(method_key)
```
banurekha: I wanted to move all the methods into private but respond_to? did not work. Looks like the other way will work.

app/mappers/datastream_json_mapper.rb
```diff
@@ -0,0 +1,271 @@
...
+      root = xml_doc.root
+      root.each_element do |element|
+        text = element.name.eql?('representative') ? format_text(element.get_text) : element.text
+        properties_array[extract_name_for(element.name)] = text
+      end
+      fedora_info['properties'] = properties_array
+    end
+
+    def content(ds)
+      content_array = {}
```
banurekha: yeah! should be properties_hash and content_properties_hash

app/mappers/datastream_json_mapper.rb
banurekha: @dbrower can we add 'ms' => 'http://www.ndltd.org/standards/metadata/etdms/1.1/', to ROF::RdfContext 

@jeremyf I can see that there is rdf (1.1.1.1 vs 2.0.1)and rubydora (1.7.4 vs 1.7.5) conflict. Should I try to

PR #569, opened 2016-11-03
Dltp 656 as librarian i want to be able

laritakr:
## Add doi for audio works

@cc0a7015dc30bcd461dde3b4316af77459c99035

## Show appropriate thumbnails based on types

@a5d279d22fdb5cccd08c72ec7b80d81c0c2dba5b

Handle thumbnails, image viewer, and audio viewer display based on work
type or file type. This added the image viewer & audio player to the
generic file's show page, as well as cleaning up the usage on the work's
show page. In addition, several unused files were deleted.

## Complete DOI functionality

@89f063d29cd8e5d151128929954c74921745619a

Finish the setup for DOI functionality on Audio objects. Make similar
changes for OSF Archive objects which also didn't function correctly.

## Remove duplicate mapping of Audio

@d0dfc028f90570251894f92a7ab6e7ec6d3de759

## Exclude OSF from DOI spec

@37226343e306c8d9e3ec3944196f280d9aac8b07

This is not implemented as we don't create an OSF Archive via the Actor.


app/views/curation_concern/base/_form_supplementary_fields.html.erb
```diff
@@ -4,7 +4,7 @@
 
 <div class="row with-headroom">
   <div class="span12">
-    <% if curation_concern.respond_to?(:identifier) %>
+    <% if curation_concern.respond_to?(:doi) %>
```
jeremyf: This looks like we will need to check other work types to see if the DOI form continues to render.

PR #574, opened 2016-11-30
Dltp 717 as the breac journal i would l

ruthtillman:
Added license to suppose Breac and standardized the licensing we already have while I was at it.


ruthtillman:
(Note, I did not change any license URIs as I know those are important, only added)

config/initializers/sufia_initializer.rb
```diff
@@ -35,18 +35,19 @@
...
-    'CC BY-ND 4.0' => 'https://creativecommons.org/licenses/by-nd/4.0/',
-    'CC BY-NC 4.0' => 'https://creativecommons.org/licenses/by-nc/4.0/',
-    'CC BY-NC-SA 4.0' => 'https://creativecommons.org/licenses/by-nc-sa/4.0/',
-    'CC BY-NC-ND 4.0' => 'https://creativecommons.org/licenses/by-nc-nd/4.0/',
-    'Attribution 3.0 United States' => 'http://creativecommons.org/licenses/by/3.0/us/',
-    'Attribution-ShareAlike 3.0 United States' => 'http://creativecommons.org/licenses/by-sa/3.0/us/',
-    'Attribution-NonCommercial 3.0 United States' => 'http://creativecommons.org/licenses/by-nc/3.0/us/',
-    'Attribution-NoDerivs 3.0 United States' => 'http://creativecommons.org/licenses/by-nd/3.0/us/',
-    'Attribution-NonCommercial-NoDerivs 3.0 United States' => 'http://creativecommons.org/licenses/by-nc-nd/3.0/us/',
-    'Attribution-NonCommercial-ShareAlike 3.0 United States' => 'http://creativecommons.org/licenses/by-nc-sa/3.0/us/',
```
jeremyf: @ruthtillman I'm double checking (as I don't recall and you probably know this off the top of your head). Are we storing URIs in the fedora objects? Or are they labels?

ruthtillman: We're storing the URIs/the => -- so we've not removed anything and when I changed it, I left http/https as is because that's in the URI that's being stored.

jeremyf:
@ruthtillman I was assuming that was the case, but I wanted to hear it from the librarians keyboard.

PR #582, opened 2017-01-05
Dltp 764 as a librarian i would like im

jeremyf:


app/repository_datastreams/image_metadata.rb
```diff
@@ -114,5 +107,141 @@ class ImageMetadata < ActiveFedora::NtriplesRDFDatastream
...
+
+    map.vra_material({to: 'material', in: RDF::VRACore}) do |index|
+    	index.as :stored_searchable
+    end
+
+    map.vra_cultural_context({to: 'culturalContext', in: RDF::VRACore}) do |index|
+    	index.as :stored_searchable
+    end
+
+    map.is_part_of ({to : 'isPartOf', in RDF::QualifiedDC}) do |index|
```
jeremyf: I believe the space betwee `map.is_part_of ({` is causing a problem. And `, in RDF` should be `, in: RDF`

app/repository_datastreams/image_metadata.rb
```diff
@@ -114,5 +107,141 @@ class ImageMetadata < ActiveFedora::NtriplesRDFDatastream
...
+    	index.as :stored_searchable
+    end
+
+    map.is_version_of ({to : 'isVersionOf' in RDF::QualifiedDC})
```
jeremyf: I believe the space betwee `map.is_part_of ({` is causing a problem. And `, in RDF` should be `, in: RDF`

ruthtillman: Ah yes this makes sense. These were the two "weird" Dublin Core things.

ruthtillman:
As Ruth's commit messages get ever more bitter. ;)

But seriously, "qualified DC" means a different thing in Curate than in most places I use it and it's important for me to remember the diff.

PR #593, opened 2017-02-01
Adjusting JSON-LD Context to not have sub-Hashes

jeremyf:
In leveraging the `JSON::LD::Writer.buffer(prefixes: CONTEXT)`, I did
some digging into how that CONTEXT is written to the buffer.

The [`JSON::LD::Context::TermDefinition`][1] is coercing the Hash into
a string (see Ruby example below).

```ruby
class JSON::LD::Context::TermDefinition
  def initialize(term, id = nil)
    @term = term
    @id = id.to_s if id # I believe this line is causing the grief
  end
end
```

Closes DLTP-751

[1]:https://github.com/ruby-rdf/json-ld/blob/1.1.4/lib/json/ld/context.rb#L35

### DIFF
--- a/app/mappers/as_jsonld_mapper.rb
+++ b/app/mappers/as_jsonld_mapper.rb
@@ -19,6 +19,11 @@ class AsJsonldMapper
   end
 
   # The keys need to be symbols to prevent the RDF library from balking.
+  # Without symbolized keys, we end up with the following exception
+  # ```console
+  #   RuntimeError:
+  #     can't add a new key into hash during iteration
+  # ```
   CONTEXT = {
       'und'.to_sym => File.join(Rails.configuration.application_root_url, 'show/'),
       'bibo'.to_sym => 'http://purl.org/ontology/bibo/',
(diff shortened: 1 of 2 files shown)


app/mappers/as_jsonld_mapper.rb
```diff
@@ -32,27 +32,27 @@ def self.call(curation_concern, **keywords)
...
-      "hasMember".to_sym => {"@id" => "frels:hasMember", "@type" => "@id" },
+      'fedora-model'.to_sym => 'info:fedora/fedora-system:def/model#',
+      'hydra'.to_sym => 'http://projecthydra.org/ns/relations#',
+      'hasModel'.to_sym => 'info:fedora/fedora-system:def/model#hasModel',
+      'hasEditor'.to_sym => 'http://projecthydra.org/ns/relations#hasEditor',
+      'hasEditorGroup'.to_sym => 'http://projecthydra.org/ns/relations#hasEditorGroup',
+      'isPartOf'.to_sym => 'info:fedora/fedora-system:def/relations-external#isPartOf',
+      'isMemberOfCollection'.to_sym => 'info:fedora/fedora-system:def/relations-external#isMemberOfCollection',
+      'isEditorOf'.to_sym => 'http://projecthydra.org/ns/relations#isEditorOf',
+      'hasMember'.to_sym => 'info:fedora/fedora-system:def/relations-external#hasMember',
```
dbrower: The `"@type" => "@id"` does make a semantic difference.

The context [in the rof repo](https://github.com/ndlib/rof/blob/master/lib/rof/rdf_context.rb) seems to work. Curious what is different in this context? using `to_sym`?

jeremyf: The issue is using the JSON::LD::Writer for the contexts; It coeces the keys of that hash into strings. I can add a hack to the post processing to leverage those contexts.

dbrower: I'm not advocating that you use those contexts. I was only wondering why the json-ld writer was turning them into strings here and not there. that is all.

Leaving out the `@type => @id` here just means the json-ld writer will insert the type when it serializes the data. so, we get the same RDF data, but a different JSON body. I have a slight preference to avoid it, but not enough to argue for it.

PR #596, opened 2017-02-08
Dltp 749 500 error on show route

banurekha:


app/views/curation_concern/generic_files/_attributes.html.erb
```diff
@@ -10,7 +10,14 @@
   <tbody>
     <tr>
       <th>Parent Work</th>
-      <td><%= link_to curation_concern.parent, polymorphic_path([:curation_concern, curation_concern.parent]) %></td>
```
jeremyf: I believe you can use:

`<%= link_to curation_concern.parent, common_object_path(curation_concern.parent.noid) %>` in all cases

Issue #606, opened 2017-03-15
change puppet script to deactivate pprd noid server


msuhovec:
see https://jira.library.nd.edu/browse/DLTP-841 for details

PR #607, opened 2017-03-15
noid-server deploy

msuhovec:
#606


msuhovec:
I tested this on libvirt6 (Remved it when I was Done).

PR #609, opened 2017-03-17
Dltp 901 curate nd figaro remediation

banurekha:


.gitignore
```diff
@@ -31,3 +31,6 @@ spec/*.zip
 spec/jetty_generator
 .tag
 .tags
+
+# Ignore application configuration
+/config/application.yml
```
jeremyf: Why are you ignoring this file? It should be included.

PR #613, opened 2017-04-26
Dltp 952 curate nd works with a represent

banurekha:
## Claiming issue DLTP-952

@be8840ee1ffa70b7185782da3b9ea46b6fdf12cc

$ ./script/start-issue DLTP-952

[skip ci]

## Fix to remove pid from representative field if fedora_object is deleted

@a05049dc26c19825f4be5c0eeeaf5c8e540ad779

Sometime fedora_objects get deleted outside the application and representative field on work have non-existent fedora object pid and causing error in application. This fix will validate if representative pid is valid if not reassign representative image pid


app/repository_models/library_collection.rb
```diff
@@ -56,6 +56,16 @@ def to_solr(solr_doc={}, opts={})
     solr_doc
   end
 
+  def check_and_clear_representative
```
jeremyf: When rendering an object for show (via a GET request), it is unexpected behavior to change the underlying object. The person viewing an object should not be able to cause the accidental side affect of updating that object. We also do not know that the next available item should be the representative. I agree with Dan's specification that:

> When a representative image is no longer present then it should render a generic thumbnail instead.

I believe the better way to handle this is to add view layer handling for `ActiveFedora::ObjectNotFoundError` exceptions. Consider extracting a helper method for rendering representative.

Also, I'm not seeing details for implementing the following:

> Also, when a GenericFile is removed we should check to see that the parent Work no longer has it set as the representative image.

banurekha: > When rendering an object for show (via a GET request), it is unexpected behavior to change the underlying object. The person viewing an object should not be able to cause the accidental side affect of updating that object. We also do not know that the next available item should be the representative. 

I can just clean up representative field and not assign representative, because there is no good way to clean this up. But then I want to have same behavior as in before_destroy hook in generic_file. Also there are some underlying assumption that work always have representative throughout the view.  

> Also, when a GenericFile is removed we should check to see that the parent Work no longer has it set as the representative image.

When GenericFile is removed through application, the check is happening that is the reason we don't see this error as often
https://github.com/ndlib/curate_nd/blob/master/app/repository_models/generic_file.rb#L20

jeremyf: I understand that there are underlying assumptions that a work always has a representative. However, when rendering a GET page, we cannot have a side-effect of rendering that page be the updating of the Fedora object. I think you'll need to rework the view logic to handle, in a non-transformative to the data manner, a work having a representative but that representative is missing.

If the image is missing from the batch ingest, I don't know what the correct response is. Perhaps @ruthtillman has some guidance - should we notify someone? Should we register an exception in Errbit but render the default image?

ruthtillman: Registering an exception but rendering the default is a reasonable behavior. Would this Errbit log ever get passed on in an easy fashion? It seems like the kind of thing which would not require immediate attention but should be looked at sooner, rather than later.

PR #614, opened 2017-04-27
Dltp 952 curate nd works with a represent

banurekha:
## Claiming issue DLTP-952

@be8840ee1ffa70b7185782da3b9ea46b6fdf12cc

$ ./script/start-issue DLTP-952

[skip ci]

## Fix to remove pid from representative field if fedora_object is deleted

@a05049dc26c19825f4be5c0eeeaf5c8e540ad779

Sometime fedora_objects get deleted outside the application and representative field on work have non-existent
fedora object pid and causing error in application. This fix will validate if representative pid is valid if not reassign representative image pid

## Reworking on presenting default thumbnail image when representative image not found in repo

@28430fe04976b6775b5c71320ca74bd62c2c6165

This fix rescue ActiveFedora::ObjectNotFoundError by presenting with default thumbnail image and reporting the error to airbrake.


app/helpers/curate_helper.rb
```diff
@@ -66,6 +66,52 @@ def curation_concern_attribute_to_html(curation_concern, method_name, label = ni
...
+    logger.error("Could not find representative pid for work: #{curation_concern.pid.inspect}")
+    Airbrake.notify(
+        error_class: exception.class, error_message: exception, parameters: { curation_concern: curation_concern }
+    )
+    markup = ''
+    markup << %(<div class="row">)
+    markup << %( <div class="work-representation span3">)
+    markup << image_tag('curate/default.png', class: "canonical-image")
+    markup << %(  </div>)
+    render_attributes(attributes_html, markup, "span9")
```
jeremyf: Is this the only place `render_attributes` is used? I'm concerned that the rather generic method name might conflict with other methods that are mixed in when helper methods are loaded.

Does this method need to exist? Can it be done inline?

config/initializers/airbrake.rb
```diff
@@ -6,6 +6,5 @@
   config.user_attributes = [:id, :username]
   config.ignore << "ActiveFedora::RecordNotFound"
   config.ignore << "ActiveRecord::RecordNotFound"
-  config.ignore << "ActiveFedora::ObjectNotFoundError"
```
jeremyf: This needs to remain. I don't want to receive messages related to this. When this is removed, everytime someone goes to a page that does not exist, we'll get an exception. I believe what needs to happen is that you need to create a new exception and send that to Airbrake instead of catching `ActiveFedora::ObjectNotFoundError` and sending that exception to Airbrake.

I'm thinking:

```ruby
# defined in lib/curate.rb
class RepresentativeObjectMissingError < RuntimeError
  attr_reader :base_exception, :work
  def initialize(base_exception, work)
    @base_exception = base_exception
    @work = work
    super("For #{work.class} ID=#{work.id}, encountered error #{base_exception}")
  end
end
```

Then you'll need to rescue the `ActiveFedora::ObjectNotFoundError` errors and create a new exception and pass that to Airbrake.

PR #634, opened 2017-08-14
Add fedora log scanner script

dbrower:
The goal is to capture how fedora is being used by all the DLT
applications, especially legacy applications.
This script is intended to be deployed by hand.
Putting it here since this appears to be the general place for
supporting scripts. (Not sure whether it belongs here or not, if there is a better place I'll put it there instead.)

### DIFF
--- a/script/scan_fedora_log.py
+++ b/script/scan_fedora_log.py
@@ -1,0 +1,54 @@
+#!/usr/bin/env python
+
+# Scan the fedora server logs and identify every fedora object requested.
+# outputs a list of each PID (with namespace) without count.
+# The environment variable SCAN_OUTPUT_DATE is used to identify which day
+# of data should be dumped.
+#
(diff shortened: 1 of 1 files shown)


dbrower:
@msuhovec yes, it is expected to set the env variable, but it is not necessary. The fedora logs seem to be rotated based on size and not on date. So the bash script rescans every log file every time. To output consistent data it seemed easiest to pass an output date to the script.

This is essentially the bash script I'm using right now
```
#!/bin/bash
filename="$home/stats-fedora-use/$(date '+%Y%m%d').csv"
# output the previous' day's usage
export SCAN_OUTPUT_DATE=$(date --date yesterday '+%Y-%m-%d')
$home/scan_fedora_log.py /opt/fedora/3.6.2/server/logs/fedora.log* | sort > $filename
```

PR #652, opened 2017-12-11
Dltp 1097 access option "Embargo" not working

laritakr:
## Claiming issue DLTP-1097

@a636f5962b2467024587c5cafa21165c140b9972

$ ./script/start-issue DLTP-1097

[skip ci]

## Render visibility selection options correctly

@b669905700e4b114f5b51acd789a29f3e7c1dab6

Visibility options returning from hydra access controls were not always
unique, resulting in the wrong radio button being checked on the edit
forms. By using AccessRenderer, the pre-checked option on the edit
form is consistent with the show views.

## Require embargo date for embargo

@aa2a93e8e2b8b0981e51d0ca8f1d63f3ce01c20c

Add javascript function to conditionally require the embargo date to be
filled in if visibility "Embargo" is selected.

## Clear embargo date for open access

@9bf85cdbe75f46e85fff70f896ec1370dec7c6c9

Use javascript function to ensure that the embargo date gets cleared
when updating the visibility on generic works to open access. (The
presence of embargo date means that the work is under embargo.)

## Use date validation on all curation_concern forms

@a576839e111d25205f8a274c4175be7c9f8a72b4

Add embargo date validate/clear function to forms for etds, patents,
senior theses, and the base form used for everything else.

## Changes to permissions partial for new works

@a9d1f916a03aa4ceeb740ae5a51a3ea7700c1e11

Need to check if the curation_concern is persisted as the permissions
radio buttons are set up.

Also documented some info regarding testing in README.md

## Closing issue DLTP-1097

@590a60dc0111106b9b57ee178286294aad3154c5

$ ./script/close-issue DLTP-1097

## Rename javascript functions

Also found and removed an extraneous function call that hadn't gotten removed when the function was moved from each radio button to the main div.


app/assets/javascripts/curate/visibility_embargo_date.js
```diff
@@ -0,0 +1,18 @@
...
+function dateRequired(source_name, object_name) {
+  if(source_name === "visibility_embargo"){
+    $('[id$=' + object_name + '_embargo_release_date]').attr('required', 'true')
+  }else{
+    $('[id$=' + object_name + '_embargo_release_date]').removeAttr('required')
+  }
+}
+
+// clear the embargo release date value for open access
+function validate(form) {
```
jeremyf: I'd prefer a more specific method name (e.g. `validateCurationConcernForm`) as `validate` is a more general than what is actually happening.

app/views/curation_concern/base/_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= simple_form_for [:curation_concern, curation_concern] do |f| %>
+<%= simple_form_for [:curation_concern, curation_concern], :html => { :onsubmit => "return validate(this);", "data-model-name" => curation_concern.class.model_name.singular } do |f| %>
```
jeremyf: See function name change recommendation above.

app/assets/javascripts/curate/visibility_embargo_date.js
```diff
@@ -0,0 +1,18 @@
+// require embargo release date for only embargo option
+function dateRequired(source_name, object_name) {
```
jeremyf: I prefer a more specific `embargoDateRequired` function name instead of the more general `dateRequired` function name. I believe it helps clarify and reduces the chance of method name collisions.

app/assets/javascripts/curate/visibility_embargo_date.js
laritakr: done

app/views/curation_concern/base/_form.html.erb
laritakr: done

app/assets/javascripts/curate/visibility_embargo_date.js
laritakr: done

PR #654, opened 2017-12-18
DLTP-1179: Fix the fixity report

jgondron:
Changes:
- The problem was due to the api changing its id key from "Id" to "ID" and the data type changing to int from string (https://github.com/ndlib/bendo/pull/167). Changed the code to match both of these changes. 
- Also fixed a new rendering problem that was causing the limit block to render over the bottom half of the form. Not sure what caused this since it appears nothing has change. Fixed the display to be grid.

### DIFF
--- a/app/views/admin/fixity/index.html.erb
+++ b/app/views/admin/fixity/index.html.erb
@@ -44,7 +44,7 @@
   <table class="table table-striped table-sortable">
     <thead>
       <tr>
-        <th v-bind:class="sortClass('')" v-on:click="changeSort('ID')">ID</th>
+        <th v-bind:class="sortClass('ID')" v-on:click="changeSort('ID')">ID</th>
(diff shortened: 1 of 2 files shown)


app/views/admin/fixity/index.html.erb
```diff
@@ -44,7 +44,7 @@
   <table class="table table-striped table-sortable">
     <thead>
       <tr>
-        <th v-bind:class="sortClass('Id')" v-on:click="changeSort('Id')">Id</th>
+        <th v-bind:class="sortClass('')" v-on:click="changeSort('ID')">ID</th>
```
jeremyf: Should this be `sortClass('ID')` instead of `sortClass('')`?

jgondron: Yes! Thanks. Not sure what happened there...

PR #660, opened 2018-02-06
Updating plural spelling for Thesis

jeremyf:
DLPT-1262

### DIFF
--- a/app/views/static_pages/home.html.erb
+++ b/app/views/static_pages/home.html.erb
@@ -160,7 +160,7 @@
         <%= image_tag("graduate_thumb.png") %>
       </div>
       <div class="collection-card-title">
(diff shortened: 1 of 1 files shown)


app/views/static_pages/home.html.erb
```diff
@@ -160,7 +160,7 @@
         <%= image_tag("graduate_thumb.png") %>
       </div>
       <div class="collection-card-title">
-        <p>Graduate School Thesis & Dissertations</p>
+        <p>Graduate School Theses & Dissertations</p>
```
dbrower: since you're in here, does erb convert `&` into `&amp;` ? just curious.

jeremyf: Probably not…I'll convert.

PR #665, opened 2018-02-22
DLTP-1261: Paginate associated files

jgondron:
## DLTP-1261: Paginate associated files

9b7c783558c08cf55676e02e0858ad51250a3062

- Added seed data for an article with many files. This meant expanding the seeds a bit to create classes and methods that would allow us to generate seed data more easily.
- Added a new method CurationConcern::WithGenericFiles#generic_files_page. This method will get a paginated result set of generic files from solr, and return both the solr response and the datastreams for the files from fedora. There didn't seem to be a natural way to do this via the built associations, so had to recreate a lot of the query/object massaging that happens in other places to get it to work with our Kaminari paginator.
- Added pagination render to the related_files partial and changed the call from generic_files (which is added via ActiveFedora has_many association) to the new paginated method
- Incidentally found a problem with how deploy runs seeds. Was looking at current path instead of release path, which means it would not use new seeds if the seeds.rb had been updated as part of this deploy.

### DIFF
--- a/app/views/curation_concern/base/_related_files.html.erb
+++ b/app/views/curation_concern/base/_related_files.html.erb
@@ -6,8 +6,8 @@
   <% if curation_concern.generic_files.present? %>
     <h2>Files</h2>
 
-    <% (@response, files) = curation_concern.generic_files_page(params[:files_page].to_i, params[:files_per_page].to_i) %>
-    
+    <% (files_solr_response, files) = curation_concern.generic_files_page(params[:files_page].to_i, params[:files_per_page].to_i) %>
+
     <% if files.any? %>
       <article class="file-delay-notice alert alert-info">
         <%= t('curate.notification_messages.file_retrieval').html_safe %>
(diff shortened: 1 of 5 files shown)


app/views/curation_concern/base/_related_files.html.erb
```diff
@@ -6,8 +6,8 @@
   <% if curation_concern.generic_files.present? %>
     <h2>Files</h2>
 
-    <% files = curation_concern.generic_files %>
-
+    <% (@response, files) = curation_concern.generic_files_page(params[:files_page].to_i, params[:files_per_page].to_i) %>
```
jeremyf: I wasn't expecting to see the setting of an instance variable in a view. Why did you choose an instance variable instead of a local variable? If this is a necessary decision to use an instance variable, please add a comment above the line explaining why.

jgondron: I would've liked to avoid this, but the underlying pagination calls expect an instance variable named response in order to work. As far as why it's here, the previous version performed the query inside of the view (the deleted line 9), so I left it here but I meant to explore moving it out. I'll take a look at that.

jeremyf: I'm okay with this being present, so long as there's an inline comment on this view that explains the decision.

PR #666, opened 2018-02-23
Hide file lists on Finding Aids

dbrower:
DO NOT MERGE

I would like to get a review on the approach. I'm waiting for final approval from archives and RBSC before I ask for a merge.

DLTP-1304


jeremyf:
Seems reasonable

PR #671, opened 2018-03-05
Bugfix deploy seeds to pprd

jgondron:
## Fix for build failing on "NoMethodError: undefined method `pp' for main:Object"

8e5b249ed5fecd98a0ba13768463e4a625eb5943

## Remove added seeds

07f7519c73978d83fdc51f5e830f8607c534ab04

- Apparently db:seed is run in production deploys as well. Pulling all of this out until I figure out a better way to handle development seeds.


laritakr:
👍

PR #679, opened 2018-04-18
Add custom development seed data

jgondron:
## DLTP-1354: Add custom development seed data

cc55f70ee9c811a640582ce20b84c6138da685b6

- These changes don't directly address 1354, but add they ability to create development seed data, which will be helpful for testing 1354
- This adds a user that has an associated Person/Profile and a user that has no associated Person/Profile in order to recreate a scenario for 1354
- Added a rake task to run the custom seeds. Ex: the seeds/dev.rb seed script can be run with `bundle exec rake db:seed:dev`

### DIFF
--- a/db/seeds/dev.rb
+++ b/db/seeds/dev.rb
@@ -50,52 +50,52 @@ create_account(user_with_profile)
 user_without_profile = create_user('userwithoutprofile@example.com', 'userwithoutprofile', 'foobarbaz')
 seeds_file = MockFile.new(__FILE__, 'text/plain', false)
 
-print "Things with no files\n"
+puts "Things with no files"
 attributes = { creator: user_with_profile.username, abstract: 'Article abstract', title: 'Article with no files' }
-print "#{attributes[:title]}\n"
+puts "#{attributes[:title]}"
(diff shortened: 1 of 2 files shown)


db/seeds/dev.rb
```diff
@@ -0,0 +1,104 @@
...
+  curation_concern
+end
+
+# Shared variables
+user_with_profile = create_user('userwithprofile@example.com', 'userwithprofile', 'foobarbaz')
+create_account(user_with_profile)
+user_without_profile = create_user('userwithoutprofile@example.com', 'userwithoutprofile', 'foobarbaz')
+seeds_file = MockFile.new(__FILE__, 'text/plain', false)
+
+print "Things with no files\n"
```
jeremyf: For reference, `puts` will add an automatic "\n" to the end of the string. The below method definition is the effective behavior of `puts`.

```ruby
def puts(string)
  print(string + "\n")
end
```

jeremyf:
Very nice!

db/seeds/dev.rb
jgondron: Ah right. So used to pretty print (pp) that when it's not available, I forget puts exists.

jeremyf: You can also do `require 'pp'` and get pp in Ruby.

PR #684, opened 2018-04-30
(DLTP-1340) Book Chapter not showing Author

jmille79:
Found and repaired issue with "Book Chapter" displaying author in search.

### DIFF
--- a/app/views/curation_concern/documents/_document.html.erb
+++ b/app/views/curation_concern/documents/_document.html.erb
@@ -13,7 +13,7 @@
     <div class="span10">
       <dl class="attribute-list">
         <%# Because Book and other Document types store primary creator information in different places. %>
-        <% if solr_doc.has?('desc_metadata__type_tesim') && Array.wrap(solr_doc['desc_metadata__type_tesim']).include?('Book') %>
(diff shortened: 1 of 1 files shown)


app/views/curation_concern/documents/_document.html.erb
```diff
@@ -13,7 +13,7 @@
     <div class="span10">
       <dl class="attribute-list">
         <%# Because Book and other Document types store primary creator information in different places. %>
-        <% if solr_doc.has?('desc_metadata__type_tesim') && Array.wrap(solr_doc['desc_metadata__type_tesim']).include?('Book') %>
+        <% if solr_doc.has?('desc_metadata__type_tesim') && (Array.wrap(solr_doc['desc_metadata__type_tesim']).include?('Book') || Array.wrap(solr_doc['desc_metadata__type_tesim']).include?('Book Chapter')) %>
```
jeremyf: I'm thinking through a way to optimize the algorithm, the duplication of `Array.wrap` and `#include?` method call, has me thinking to replace it with:

```ruby
if solr_doc.has?('desc_metadata__type_tesim') && Array.wrap(solr_doc['desc_metadata__type_tesim']).detect { |type| type == 'Book' || type == 'Book Chapter' }
```

This means we only convert the object once to an array, then break out of the loop once we have a hit.

PR #687, opened 2018-04-30
Create the Person/Profile for new users

jgondron:
## DLTP-1354: Create the Person/Profile for new users

68045e8c4b08c54d81c000b2103ff49fe3d72e89

- Added a call to create the Person and Profile for an authenticated user if one does not exist, using the net id as the name. This should allow assigning the user as an editor of a work as long as they've logged in.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -34,6 +34,7 @@ class ApplicationController < ActionController::Base
   end
 
   def after_sign_in_path_for(resource)
+    ensure_user_has_profile(resource) if resource.is_a?(User)
     session[:previous_url] || root_path
   end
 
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -50,4 +51,14 @@ def new_session_path(scope)
     send("new_#{scope}_session_path")
   end
   helper_method :new_session_path
+
+  # Creates an associated Person and Profile if none exists after the User logs in.
+  # This is a stop gap to handle https://jira.library.nd.edu/browse/DLTP-1354. We should
+  # remove this once we remove Person and/or Profile objects.
+  def ensure_user_has_profile(user)
+    unless user.person && user.person.profile.present?
```
jeremyf: I find `unless` with `&&` a bit hard to mentally parse. I believe the following are logically the same. Does this improve legibility for others?

```
return true if user.person && user.person.profile.present?
account = Account.to_adapter.get!(user.id)
update_status = account.update_with_password({ "name" => user.username })
```

I'm neither approving nor requesting changes, just looking for a conversation.

jgondron: I usually stray from it for more complex bits of logic, especially when there's a lot of nested logic. But for simple cases like this, I find it reads ok. But I'll let others chime in and change if needed.

jmille79: Both are easy to follow. Since it is a simple case and intent is commented, I see no reason to change it.

PR #697, opened 2018-05-10
DLTP-1258: Sort missing to bottom

jgondron:
## DLTP-1258: Sort missing to bottom

e3adbea324269c2a3898abbbca5a1cef802fd4e4

This can be done at query time using something like 'sort=if(ord(date_created_derived_dtsim),ord(date_created_derived_dtsim),999999) asc', but I decided to do it at the schema level. This required a number of changes:
- Change the schema for the _dtsi field to sortMissingLast
- Change the model to use a single value date field. Which meant changing the params passed to solrizer's create_and_insert_terms method to :stored_sortable, and passing a single date instead of an array of strings (otherwise it would not recognize it as a date field and wouldn't add the _dt suffix
- Changed the catalog controller to match, and no longer need to use ord function since we can sort on a single value date field naturally.
- This also meant fixing a problem with the Dockerfile. Was previously just using the downloaded schema/configs. Had to add a copy command to make sure that the schema used in dev/test were up to date with what's in the curate repo.

### DIFF
--- a/app/repository_models/curation_concern/model.rb
+++ b/app/repository_models/curation_concern/model.rb
@@ -60,6 +60,8 @@ protected
         parse_date(publication_date)
       when self.respond_to?(:date_issued)
         parse_date(date_issued)
+      when self.respond_to?(:date)
+        parse_date(date)
       when self.respond_to?(:date_created)
         parse_date(date_created)
       end
--- a/app/repository_models/curation_concern/model.rb
+++ b/app/repository_models/curation_concern/model.rb
@@ -55,22 +55,19 @@ protected
       # so I'm putting this here. If this changes, then it may make sense to move
       # this logic into the individual model's to_solr methods.
       # See ticket DLTP-1258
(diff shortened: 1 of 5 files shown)


app/repository_models/curation_concern/model.rb
```diff
@@ -55,22 +55,19 @@ def add_derived_date_created(solr_doc)
...
+        parse_date(date_issued)
       when self.respond_to?(:date_created)
-        parse_dates(date_created)
+        parse_date(date_created)
       end
-      self.class.create_and_insert_terms('date_created_derived', derived_dates, [:dateable], solr_doc)
+      self.class.create_and_insert_terms('date_created_derived', derived_date, [:stored_sortable], solr_doc)
     end
 
-    def parse_dates(source_dates)
```
jeremyf: Wondering about the shift from many to one date to parse. Are all of these field values stored in Fedora singular?

PR #708, opened 2018-06-12
Changed Language

jmille79:
Per Request, 'Upload a File' is now 'Upload Files' and the file size warning has been removed.


app/helpers/curate_helper.rb
```diff
@@ -3,10 +3,6 @@ def support_email_link(options = {})
     mail_to(t('sufia.help_email'), t('sufia.help_email'), { subject: t('sufia.help_email_subject') }.merge(options)).html_safe
   end
 
-  def file_size_warning
```
jgondron: Why was this removed?

app/views/curation_concern/datasets/_form_files_and_links.html.erb
```diff
@@ -24,10 +24,8 @@
 <% else %>
   <div class="row">
     <fieldset class="span5">
-      <legend>Upload a File</legend>
-
-      <%= render 'file_size_warning' %>
-      <%= f.input :files, as: :file %>
+      <legend>Upload Files</legend>
+      <%= f.input :files, as: :file, input_html: {multiple: true} %>
```
jgondron: I would expect this language, and the multiple: true, to only be on the views you've already changed. Is this enabling users to upload multiples on update, even though the underlying controllers/actors won't work yet? Or were these just missed in the previous PR?

jmille79: Sorry about that, overlapped with 1431. Cleaning it up now.

app/helpers/curate_helper.rb
jmille79: Don had asked for the File Size Warning to be removed on the upload section. It was the only code using that, so I removed it from the helper as well.

dbrower: I meant for the words "Please Note:" to be removed. Does this only remove those words? or the entire warning?

PR #709, opened 2018-06-12
Multi Upload of files after creation

jmille79:
DLTP-1341, Restructured the file creation to allow for multiple files to be uploaded at once. Adjusted the input options to reflect.


app/services/curation_concern/generic_file_actor.rb
```diff
@@ -2,7 +2,16 @@ module CurationConcern
   class GenericFileActor < CurationConcern::BaseActor
 
     def create
-      super { update_file  && download_create_cloud_resources }
+      # Get the list of files, then for each file, go through the creation process
+      files = attributes.delete(:file).to_a
+      times = 0
+      while times < files.length do
```
jgondron: Would prefer for each here, see https://ruby-doc.org/core-2.2.0/Array.html#method-i-each

app/services/curation_concern/generic_file_actor.rb
```diff
@@ -2,7 +2,16 @@ module CurationConcern
...
+        @file = files[times]
+        attach_files && download_create_cloud_resources &&
+        apply_access_permissions
```
jgondron: Either indent this, or continue it on line 10. This is still part of that conditional

app/services/curation_concern/generic_file_actor.rb
```diff
@@ -36,6 +45,46 @@ def update_file
       end
     end
 
+    def attach_files
```
jgondron: Nitpicky, but this should probably be named "attach_file" since it's only interacting with the current @file.

app/services/curation_concern/generic_file_actor.rb
jmille79: Old habits... All changes applied

PR #712, opened 2018-07-05
Dltp 1202

msuhovec:


app/services/fedora_tools.rb
```diff
@@ -64,7 +64,7 @@ def self.get_works_pid_list(input_list)
     output_list = {}
    
     input_list.each do |pid, record|
-      output_list[pid]= pid if is_work( record.models[record.models.length - 1])
+	    output_list[pid]= pid.gsub("und:","") if is_work( record.models[record.models.length - 1])
```
dbrower: could this use `#last()`?

Should the code assume the work model is the last item in the list? maybe, use `.any?` instead?

PR #713, opened 2018-07-09
DLTP-1419 Update Etd Fields

jmille79:
Script designed to update Fedora data of sipity objects who did not receive the data at ingest

### DIFF
--- a/script/update_etd_fields.rb
+++ b/script/update_etd_fields.rb
@@ -1,0 +1,67 @@
+require 'json'
+# Make sure a file is given
+if ARGV.length < 1
+  puts "ERROR: No File listed in command
+  usage: update_empty_etd_fields.rb update_info.json
+
+  json example:
+  [
+    {
+      'work_id': '00000001a1',
+      'catalog_system_number': '000000000',
+      'ETD_REVIEWER_SIGNOFF_DATE': '2010-01-01',
+      'oclc_number': 000000000
+    }
+  ]"
+  exit 2
+end
+# Read JSON file
+puts 'Reading JSON File'
+filename = ARGV[0]
+file_content = File.read(filename)
+data = JSON.parse(file_content)
+
+# typemap = {
+#   'work_id' => 'pid',
+#   'catalog_system_number' => 'alephIdentifier',
+#   'ETD_REVIEWER_SIGNOFF_DATE' => 'date_approved',
+#   'oclc_number' => 'urn'
+# }
+
+data.each do |record|
+# Check if the record exists in CurateND
+  puts "checking record #{record['work_id']}"
+  # next puts "no record found, continuing to next record" unless Etd.exists?(record['work_id'])
+  begin
+    etd = Etd.find('und:'+record['work_id'])
+  rescue
+    next puts "no record found, continuing to next record"
+  end
+  record_updated = false
+# Check if the record has the attributes set
+  unless record['catalog_system_number'].nil?
+    etd.alephIdentifier = record['catalog_system_number']
+    record_updated = true
+    puts "new record found"
+  end
+  unless record['ETD_REVIEWER_SIGNOFF_DATE'].nil?
+    etd.date_approved = record['ETD_REVIEWER_SIGNOFF_DATE']
+    record_updated = true
+    puts "new record found"
+  end
+  unless record['oclc_number'].nil?
+    etd.urn = record['oclc_number']
+    record_updated = true
+    puts "new record found"
+  end
+# Save etd if any records updated
+  if record_updated
+    begin
+      puts "saving record"
+      etd.save!
+      puts "record saved"
+    rescue
+      puts "record #{record['work_id']} failed to save"
+    end
+  end
+end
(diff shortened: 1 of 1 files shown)


script/update_etd_fields.rb
```diff
@@ -0,0 +1,67 @@
...
+    record_updated = true
+    puts "new record found"
+  end
+  unless record['oclc_number'].nil?
+    etd.urn = record['oclc_number']
+    record_updated = true
+    puts "new record found"
+  end
+# Save etd if any records updated
+  if record_updated
```
jgondron: Can you test if the object was changed before saving and output the id with the new values? This would be nice to log to a file as a means for auditing what objects were changed as part of this remediation. I'm not sure if these models implement this correctly, but ActiveRecord usually provides a changed? method https://apidock.com/rails/ActiveRecord/Dirty/changed%3F that might help with this.

script/update_etd_fields.rb
```diff
@@ -0,0 +1,67 @@
...
+      'work_id': '00000001a1',
+      'catalog_system_number': '000000000',
+      'ETD_REVIEWER_SIGNOFF_DATE': '2010-01-01',
+      'oclc_number': 000000000
+    }
+  ]"
+  exit 2
+end
+# Read JSON file
+puts 'Reading JSON File'
```
dbrower: If you are `puts`ing that you are reading a file, might as well also print the name of the file. That will help with diagnostics.

script/update_etd_fields.rb
```diff
@@ -0,0 +1,67 @@
...
+  puts "checking record #{record['work_id']}"
+  # next puts "no record found, continuing to next record" unless Etd.exists?(record['work_id'])
+  begin
+    etd = Etd.find('und:'+record['work_id'])
+  rescue
+    next puts "no record found, continuing to next record"
+  end
+  record_updated = false
+# Check if the record has the attributes set
+  unless record['catalog_system_number'].nil?
```
dbrower: Is the sipity data the source of record? It seems like we should first check that the field in the curate record is blank before overwriting it with the sipity stuff, no?

script/update_etd_fields.rb
```diff
@@ -0,0 +1,67 @@
...
+    etd.alephIdentifier = record['catalog_system_number']
+    record_updated = true
+    puts "new record found"
+  end
+  unless record['ETD_REVIEWER_SIGNOFF_DATE'].nil?
```
dbrower: ditto above comment.

script/update_etd_fields.rb
```diff
@@ -0,0 +1,67 @@
...
+    etd.date_approved = record['ETD_REVIEWER_SIGNOFF_DATE']
+    record_updated = true
+    puts "new record found"
+  end
+  unless record['oclc_number'].nil?
```
dbrower: ditto above comment.

script/update_etd_fields.rb
```diff
@@ -0,0 +1,67 @@
...
+    etd = Etd.find('und:'+record['work_id'])
+  rescue
+    next puts "no record found, continuing to next record"
+  end
+  record_updated = false
+# Check if the record has the attributes set
+  unless record['catalog_system_number'].nil?
+    etd.alephIdentifier = record['catalog_system_number']
+    record_updated = true
+    puts "new record found"
```
dbrower: This puts does not look useful since a new record is not found...it is already in curate. also this is the exact same output from each of the three conditionals. If each conditional had its own unique phrase, debugging could follow which fields were changed.

script/update_etd_fields.rb
```diff
@@ -0,0 +1,74 @@
...
+  @logger.info "checking record #{record['work_id']} for #{record['work_id']}"
+  # next @logger.info "no record found, continuing to next record" unless Etd.exists?(record['work_id'])
+  begin
+    etd = Etd.find('und:'+record['work_id'])
+  rescue
+    next @logger.info "no record found, continuing to next record"
+  end
+  record_updated = false
+# Check if the record has the attributes set
+  unless record['catalog_system_number'].nil? || !etd.alephIdentifier.nil?
```
jgondron: This is getting into territory where I would prefer if over unless. You can also likely replace !nil? with a presence check: https://apidock.com/rails/Object/present%3F. This would also make it ingest the value if the property in Fedora is there, but as an empty string, where as nil? only checks explicitly for nil. So either change to something like:

`unless record['catalog_system_number'].nil? || etd.alephIdentifier.present?`
or 
`if record['catalog_system_number'].present? && etd.alephIdentifier.blank?`

Would prefer the latter. Just double check my logic :)

script/update_etd_fields.rb
```diff
@@ -0,0 +1,74 @@
...
+# typemap = {
+#   'work_id' => 'pid',
+#   'catalog_system_number' => 'alephIdentifier',
+#   'ETD_REVIEWER_SIGNOFF_DATE' => 'date_approved',
+#   'oclc_number' => 'urn'
+# }
+
+data.each do |record|
+# Check if the record exists in CurateND
+  @logger.info "checking record #{record['work_id']} for #{record['work_id']}"
```
dbrower: Did you intend to log the `work_id` twice here?

PR #717, opened 2018-07-18
DLTP-1459 Frontend changes

jmille79:
Changed front-end settings to allow for ldap search of netid


jmille79:
Merged front end into back end for testing

PR #721, opened 2018-07-23
Fix "mysql Incorrect string value: '\xE2\x9C\x93"\x0As" error

jgondron:
## Fix "mysql Incorrect string value: '\xE2\x9C\x93"\x0As" error

3de2ec2809f50d2aba51ce29afe3b7f0a22e999f

Problem was that the default character set in mariadb was not being set to utf8. Added a custom config to the docker compose that will set this every time mariadb starts


jmille79:
Testede and working

PR #725, opened 2018-07-27
Dltp 1465

jmille79:


app/helpers/curate/collections_helper.rb
```diff
@@ -7,10 +7,6 @@ def button_for_create_new_collection(label = 'Create Collection')
     render partial: 'button_create_collection', locals:{label:label}
   end
 
-  def title_for_new_form(profile_section)
```
jgondron: Just wondering why this was removed. Would have expected to see a call to this method also removed, but don't see any in the rest of the changes. Are you sure this doesn't break anything?

h-parekh:
@jmille79 Consider giving a more generic title to PRs that span across different jira tickets. Additionally, you could use [commandline-tools](https://github.com/ndlib/commandline-tools/blob/master/bin/build-pull-request-message) to build PR descriptions.

PR #739, opened 2018-09-04
Removing "My Collections" button

jeremyf:
Per DLTP-1537:

> the 'Manage > My Collection' button is creating confusion amongst
> users that they can create their own collections.
>
> Per Don Brower, the user collection functionality was removed two
> years ago, and so this button should be removed.


jeremyf:
@dbrower if you would prefer to view this in the browser, let me know. Otherwise, due to the simplicity of this update, I'll consider this approval to be UA complete.

dbrower:
perfect!
LGTM.

PR #742, opened 2018-09-06
Wordsmithing

jeremyf:
Per DLTP-1539, updating the language to explain starred/required
fields.


jeremyf:
@dbrower The generic_work's base form is "inherited" by downstream implementations

Issue #201, opened 2015-07-06
Add standard DOI formatting to DOI field in CurateND


cdeluca495:
CurateND does not enforce a standard format of DOI data entry.  The values entered into CurateND should adhere to the DOI format as outlined here. 
http://www.doi.org/factsheets/DOIIdentifierSpecs.html

jeremyf:
From the DOI Specification:

> A DOI name may be represented as a URL (http string) by prefacing the string with http://doi.org/ (or http://dx.doi.org) to the DOI of the document (e.g., to resolve the DOI name 10.1000/182, enter into a browser the address: http://doi.org/10.1000/182). Web pages or other hypertext documents can include hypertext links in this form.

So our sanitization can strip any doi.org or dx.doi.org prefix. Below is what I believe to be regular expression to sanitize.

```ruby
%r{^(https?://)?(dx.)?doi.org/}
```

jeremyf:
Don't spend time working on validating the format; Instead spend time stripping the possible prefixes for domains.

dbrower:
I think a better regexp is along the line of s/^.*10\./10./ that is, strip everything up to the initial "10.", and then make sure it starts with "10.".

danhorst:
Example record: `und:mk61rf58938` https://curate.nd.edu/show/mk61rf58938

dbrower:
I went to the example record and I do not see a DOI. It does not appear on the show page.

ruthtillman:
Currently, DOIs aren't displaying at all records at all. For example, I made: https://libvirt8.library.nd.edu/concern/articles/tx31qf88686 

I'm not seeing it in https://github.com/ndlib/curate_nd/blob/master/app/views/curation_concern/articles/_attributes.html.erb etc.

Metadata includes a DOI.

<info:fedora/und:tx31qf88686> <http://purl.org/dc/terms/dateSubmitted> "2016-05-24Z"^^<http://www.w3.org/2001/XMLSchema#date> .
<info:fedora/und:tx31qf88686> <http://purl.org/dc/terms/title> "DOI Tester document" .
<info:fedora/und:tx31qf88686> <http://purl.org/dc/terms/creator#author> "Ruth Tillman" .
<info:fedora/und:tx31qf88686> <http://purl.org/dc/terms/abstract> "Yadda Yadda, oh c'mon that's not a typo." .
<info:fedora/und:tx31qf88686> <http://purl.org/dc/terms/creator#affiliation> "" .
[...]

PR #746, opened 2018-09-07
Refining Doi::Datacite.normalize_identifier

jeremyf:
## Adding testing tables

2a3e70e495172e63711caa3253374cb85c2f5e14

Instead of the chatty individual rspec scenarios, compress those into a
test tables. This helps clarify at a glance what is being tested.

## Refining Doi::Datacite.normalize_identifier

64fb24fa753b262d90c96d5efbe2abfc11b42a5a

Based on the following comment in JIRA, the solution is to ensure that
past scenarios continue to work when we apply different implementations
of the normalize_identifier.

> We've seen this before: https://github.com/ndlib/curate_nd/issues/201
>
> As noted there, a DOI is not a URL. Also, there appears to be no
> standard for prefixing DOIs. I think the best normalization routine is
> to look for the initial "10." in the DOI, strip of everything before
> and prefix it with "doi:".
>
> DOIs in that format seem to resolve ok, since the dx.doi.org resolver
> can deal with a "doi:" prefix.

DLTP-1528

### DIFF
--- a/app/services/doi/datacite.rb
+++ b/app/services/doi/datacite.rb
@@ -7,11 +7,8 @@ module Doi
 
     def self.normalize_identifier(value)
       value.to_s.strip.
-        sub(/\A#{resolver_url}\/*/, '').
-        sub(/\A\s*doi:\s+/, 'doi:').
-        sub(/\A(\d.*)/, 'doi:\1').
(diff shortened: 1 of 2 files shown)


dbrower:
This happened while i was on vacation, but +1 to converting the tests into a table. MUCH more understandable.

PR #747, opened 2018-09-07
Ensuring expanded ToS allows acceptance

jeremyf:
## Ensuring expanded ToS allows acceptance

49767a761b59d836ece78c57adc6d8a8997f0366

Prior to this commit, the HTML modal was rendering inside the form.
This would work, except that the modal JS would pull that HTML
and render the DOM elements outside of the form. This resulted in a
loss of the form context, and thus the submit buttons didn't know
where to submit the accept or decline action.

By moving the modal content outside of the form, and adding a form
around the submit buttons:

* the expected behavior is restored
* we don't violate HTML standards by having a form nested within
  another form

In addition, I've added a confirmation flash message to let the end
user know that the ToS was accepted.

DLTP-1521

## Updating redirect for ToS agreement

5b7756ff44b5686a6317f4ba80cec039f6e03c85

> The standard use case for users accepting ToS is that they are
> entering the system for the first time. As such, the next likely step
> is filling out their registration information.

DLTP-1521

### DIFF
--- a/app/controllers/terms_of_service_agreements_controller.rb
+++ b/app/controllers/terms_of_service_agreements_controller.rb
@@ -11,6 +11,7 @@ class TermsOfServiceAgreementsController < ApplicationController
   def create
     if user_just_agreed_to_tos?
       current_user.agree_to_terms_of_service!
+      flash[:notice] = "You've agreed to the Terms of Service."
       redirect_to new_classify_concern_path
     else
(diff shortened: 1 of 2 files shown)


h-parekh:
Thanks for the detailed commit message @jeremyf. Confirmation flash message is a nice touch!

PR #749, opened 2018-09-19
Fix Reindex collections

laritakr:
Fixes https://jira.library.nd.edu/browse/DLTP-1550 and
https://jira.library.nd.edu/browse/DLTP-1513

Issue 1550 was caused by PR https://github.com/ndlib/curate_nd/pull/736
which was reversed with PR https://github.com/ndlib/curate_nd/pull/744

This changes the fix for DTLP-1513 without causing the problems which caused DLTP-1550.

**Changes**
- Add nested Library Collections to seed data for testing
- Remove duplicate reindexing from reindex_worker. Calling update_index already does the nested indexing.
- Fix raw_child_documents_of method. It didn't work when the solr_doc existed but didn't have the nested indexing fields in it. It also found only the default number of rows rather than the full set of children.

### DIFF
--- a/lib/curate/library_collection_indexing_adapter.rb
+++ b/lib/curate/library_collection_indexing_adapter.rb
@@ -55,14 +55,17 @@ module Curate
     # @param curation_concern_type [nil, ActiveFedora::Base] filter on this and only this type of curation concern
     # @return [Hash] A raw response document from SOLR
     def self.raw_child_documents_of(parent_document, curation_concern_type = nil)
-      type_query = "_query_:\"{!raw f=has_model_ssim}info:fedora/afmodel:#{curation_concern_type}\""
-      # Need to find all documents that have ancestors equal to one or more of the given parent_document's pathnames
-      pathname_query = parent_document.pathnames.map do |pathname|
-        text = "_query_:\"{!raw f=#{SOLR_KEY_ANCESTOR_SYMBOLS}}#{pathname.gsub('"', '\"')}\""
-        text += " AND #{type_query}" if curation_concern_type
-        text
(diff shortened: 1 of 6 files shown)


lib/curate/library_collection_indexing_adapter.rb
```diff
@@ -55,14 +55,17 @@ def self.each_child_document_of(parent_document, curation_concern_type = nil, &b
...
-        text = "_query_:\"{!raw f=#{SOLR_KEY_ANCESTOR_SYMBOLS}}#{pathname.gsub('"', '\"')}\""
-        text += " AND #{type_query}" if curation_concern_type
-        text
-      end.join(" OR ")
-      ActiveFedora::SolrService.query(pathname_query)
+      # gather the list of id's to load solr documents for
+      pid_list = []
+      parent_fedora_object = ActiveFedora::Base.find(parent_document.pid, cast: true)
+      pid_list = parent_fedora_object.library_collection_member_ids if parent_fedora_object.class == LibraryCollection
+
```
jeremyf: If the resulting pid_list is empty, can we short-circuit out of here?

laritakr: Not easily.

First of all, the pid_list will only be empty if for a library collection that has no subcollections and no member objects. There aren't a lot of these situations... usually when we create a collection, we put things into it.

Secondly, places calling this method expect a raw Solr response as the returned value. Short-circuiting out bypasses the expected type of returned value, meaning we would need to handle alternate types of responses where this method is called. It was my feeling that a solr response should always be returned, even if there are no docs in it.

PR #751, opened 2018-10-03
Get all members of Library collection

laritakr:
refers to https://github.com/ndlib/curate_nd/pull/750

Perform a second solr query to get all of the members of a collection.
(prior attempt to read through and accumulate all of the pages was
not loading new pages reliably.)


lib/curate/library_collection_indexing_adapter.rb
```diff
@@ -59,21 +59,16 @@ def self.raw_child_documents_of(parent_document, curation_concern_type = nil)
...
-      full_response = solr_response['response']['docs']
-
-      loop do
-        total_read += rows
-        break if solr_response['response']['numFound'] <= total_read
-        page += 1
-        solr_response = get_page_from_solr(qry: qry, params: { fq: fq, rows: rows, page: page } )
-        full_response += solr_response['response']['docs']
+
+      solr_response = get_page_from_solr(qry: qry, params: { fq: fq, rows: rows } )
```
jeremyf: If I'm reading the intention of this, and perhaps comments would help, the `rows = 12` is a bit of a concession saying "Most collections will have less than 12 documents (or less)", so use that query. However, if that is not the case, use the actual number of documents.

I can get behind that.

laritakr: I added some documentation. Might it be better to use a larger number so there are less times we have to re-query to get the full number? 

I think something like 50 would catch at least half, and 100 would catch the majority of the collections we currently have.

PR #768, opened 2019-01-04
Dltp 1518 allow editors to create links

laritakr:
Issue https://jira.library.nd.edu/browse/DLTP-1518

## Temporary Access Tokens
* Allow creation of tokens by anyone with edit rights, via a link from each work's page.
* Combine show view for access token into edit views. Remove show view.
* Add alternate views of access token list, limited to only the tokens someone is allowed to manage, and optionally all of the tokens for one file ID.
* Fix ability logic for token manager (it didn't actually work before!)
* Add edit-checking to creation of tokens and customize error messages.
* Add token list option to "manage" menu.
* Add option for Token Manager to remove all expired or out-dated access tokens
## Admin menu changes
* Replace errbit link with sentry link on admin menu.
* Moved token list off of admin menu (and onto manage menu instead).
## Requests for Access
* Allows for more than just ETD's to show the access request option, based on content of access_request_map.yml file. This is a first step toward a future option to allow access requests based on a metadata field on a work.


app/controllers/admin/temporary_access_tokens_controller.rb
```diff
@@ -64,4 +83,35 @@ def new_temporary_access_token_params
...
+  end
+
+  def validate_create_token
+    item_to_access = begin
+      ActiveFedora::Base.load_instance_from_solr(Sufia::Noid.namespaceize(@temporary_access_token.noid))
+    rescue ActiveFedora::ObjectNotFoundError
+      return { test: :not_found, valid: false, notice: "Error: Unable to add access token; file #{@temporary_access_token.noid} was not found." }
+    end
+    # object must be a generic_file
+    return { valid: false, notice: "Error: Unable to add access token; item ID is not a file." } if item_to_access.class != GenericFile
```
jeremyf: Prefer `item_to_access.is_a?(GenericFile)`

laritakr: done

jeremyf:
@laritakr Is this ready for a review and merge?

laritakr:
@jeremyf not yet... @dbrower is reviewing the functionality with Mikala and Adam today to see if they want any additional changes.

app/controllers/temporary_access_tokens_controller.rb
```diff
@@ -0,0 +1,131 @@
...
+    return { valid: false, notice: "Unable to add access token: item #{@temporary_access_token.noid} is not a file." } unless item_to_access.is_a? GenericFile
+    # must be allowed to edit file or token manager
+    return { valid: false, notice: "Error: Not authorized to create token for file #{@temporary_access_token.noid}." } unless (can? :edit, item_to_access) || (can? :manage, TemporaryAccessToken)
+    # otherwise valid
+    { valid: true, notice: nil }
+  end
+
+  private
+
+  def load_limited_item
```
jeremyf: I'm okay with this, but there is an assumption when calling load_limited_item that `@limit_to_id` is set. You could pass that value of @limit_to_id as a parameter. Not a blocker, just a consideration if you are in the code.

app/helpers/curation_concern_helper.rb
```diff
@@ -17,4 +17,37 @@ def decode_administrative_unit(curation_concern, method_name, label = nil, optio
...
+    # return some metadata field if this_request["access_request_method"] == "metadata"
+    nil
+  end
+
+  def this_request
+    @this_request ||= access_request_data[curation_concern.class.to_s.downcase]
+  end
+
+  def access_request_data
+    TemporaryAccessToken.access_request_data
```
jeremyf: I'm a bit surprised that this works. The `TemporaryAccessToken.access_request_data` would be calling a class method on `TemporaryAccessToken`. However, in the implementation I see that `access_request_data` is an instance method (e.g. you'd need to do TemporaryAccessToken.new.access_request_data`. My reading of the method definition is that this could be a class method. Which would mean you'd need to change the method usage in the model.

app/models/temporary_access_token.rb
```diff
@@ -61,4 +61,72 @@ def strip_pid_namespace
...
+  def token_file
+    @token_file ||= ActiveFedora::Base.load_instance_from_solr(Sufia::Noid.namespaceize(noid))
+  end
+
+  def parent_class
+    @parent_class ||= token_file.parent.class.to_s.downcase
+  end
+
+  # Data file controlling all access requests
+  def access_request_data
```
jeremyf: I believe this should be a class method (e.g. `def self.access_request_data`). Which would mean on line 117, you'd want to say `self.class.access_request_data[parent_class]`

app/models/temporary_access_token.rb
```diff
@@ -61,4 +61,72 @@ def strip_pid_namespace
...
+
+  def this_request
+    @this_request ||= access_request_data[parent_class]
+  end
+
+  def token_file
+    @token_file ||= ActiveFedora::Base.load_instance_from_solr(Sufia::Noid.namespaceize(noid))
+  end
+
+  def parent_class
```
jeremyf: Perhaps consider naming this `parent_class_name`; In my reading I would assume `parent_class` to be a class (e.g. `GenericWork` instead of `"genericwork"`

app/helpers/curation_concern_helper.rb
laritakr: I thought I had it as a class method! I'm not sure why it worked locally, but it didn't work on staging and I hadn't figured it out yet.

app/controllers/temporary_access_tokens_controller.rb
laritakr: I have to check whether limit_to_id is nil at some point because it's not always loaded. I thought it made sense to not call the load method at all when you don't have the id, rather than passing it in and then testing whether the id is empty within the method. So I chose to use `@limited_item = load_limited_item unless @limit_to_id.nil?`, which guarantees that the id is present when you get into the method.

I could change this, but would like to understand why it's better to pass it in.

app/controllers/temporary_access_tokens_controller.rb
```diff
@@ -0,0 +1,131 @@
...
+      {}
+    end
+  end
+
+  # Enforce that the current user is associated with creating or modifying a token
+  def temporary_access_token_params_with_current_user
+    temporary_access_token_params.merge({ :issued_by => current_user.user_key })
+  end
+
+  def temporary_access_token_list
```
jeremyf: This is a non-blocker, but one thing you can do is:

```ruby
finding_scope = TemporaryAccessToken.order(updated_at: :desc).page(params[:page])
finding_scope = finding_scope.where(noid: @limit_to_id) if @limit_to_id.present?
finding_scope
```

This can then ensure that you keep the two branches of the if statement the same.

app/models/temporary_access_token.rb
```diff
@@ -61,4 +61,72 @@ def strip_pid_namespace
...
+  end
+
+  def this_request
+    @this_request ||= self.class.access_request_data[parent_class_name]
+  end
+
+  def token_file
+    @token_file ||= ActiveFedora::Base.load_instance_from_solr(Sufia::Noid.namespaceize(noid))
+  end
+``
```
jeremyf: I think these need to be removed. (as written the double back ticks will open a shell in the application, that does nothing, and then close the shell)

laritakr: ok wow I have no idea how those got in there!

PR #776, opened 2019-03-12
Removing "logger" gem

jeremyf:
Based on a report, on a January 9th deploy, we began noticing
slow-downs. The one change was a deploy and locking logger to this
application.


dbrower:
This gem can be removed with no code changes?

jeremyf:
@dbrower it would appear so. I believe it implemented a common interface for loggers, and anywhere that had `require "logger"` would reference the gem instead of the upstream system logger (or Rails logger). It also appears that it was used for some ETD migration work.

However, all of this is a stab in the dark at brining back some performance.

PR #778, opened 2019-03-25
Removing app/models/generic_file.rb

jeremyf:
This was copied over from sufia, but we have an overlay in the following
location:

```console
./app/repository_models/generic_file.rb
```

### DIFF
--- a/app/models/generic_file.rb
+++ b/app/models/generic_file.rb
@@ -1,3 +1,0 @@
-class GenericFile < ActiveFedora::Base
-  include Sufia::GenericFile
-end


jeremyf:
Steps to test fix:
```console
ssh to-PPRD
cd curatend/current
bundle exec rails console pre_production
ActiveFedora::Base.find("und:9019s18110b", cast: true).with_empty_content?
=> false
```
_Note: the current directory is the deployed version of this code._
```console
ssh to-PPRD
cd curatend/releases/20190313102434
bundle exec rails console pre_production
ActiveFedora::Base.find("und:9019s18110b", cast: true).with_empty_content?
=> NoMethodError: undefined method `with_empty_content?' for #<ActiveFedora::Base pid:"und:9019s18110b", >
```
_Note: the `releases/20190313102434` directory is the previously deployed code base._

PR #780, opened 2019-04-01
Update FAQ

dbrower:
Reorder questions. Revise answers. Add more questions. Add link to OSF.

There are no code changes in this PR.

### DIFF
--- a/app/views/static_pages/faqs.html.erb
+++ b/app/views/static_pages/faqs.html.erb
@@ -15,7 +15,7 @@
   </h1>
 
   <h2>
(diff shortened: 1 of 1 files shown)


mikalarae:
@dbrower is this where/how I comment? I think so. This looks really great! I only had a few nitpick-y things (below) which do not relate to the content.

Line 71: Instead of "won't," perhaps "will not"
Also Line 71: ... "content will be available _and preserved_ even after you leave the University"
Line 90: instead of "don't," perhaps "do not"
Line 130: maybe "Do I _relinquish_ my rights"

Again, all small changes that don't affect the content. Let me know what you think!

PR #789, opened 2019-05-07
Update docker build

jgondron:
Several changes have been made to things like ruby versions and libraries. This brings the docker build up to speed with those changes.
- Changed rails image to pull ruby 2.3.8 for base image
- Added libssl1.0-dev to the rails image due to a strange segfault issue when using thin on the container
- Removed the lib/logger copy from the rails image, since this is no longer in the repo
- Additionally separated out the compose files to give us options on how to run this in development, since some users would like to use it for running dependencies only (mysql/fedora/solr). Updated the readme to describe how to run it in the different configurations

As an aside, as part of trying to figure out the segfault issue with thin, I tried both puma and passenger. Both are relatively easy to convert to. Puma is twice as slow with rendering the home page. Passenger takes an enormous hit on first load, but then is faster to load the home page. So, left it using thin, but there may be room for experimenting with this a bit more in the future to more closely match dev to how Curate will run in production.

### DIFF
--- a/README_DOCKER.md
+++ b/README_DOCKER.md
@@ -5,54 +5,78 @@ Install Docker: https://www.docker.com/
 
 Install docker-sync: http://docker-sync.io/
 
-Start the stack:
+## Starting services
+### Dependencies only
+If you want to run the dependencies without running rails, such as when running rails on your host machine, just run the base compose:
+```console
+docker-compose up
+```
+Then load the database schema:
+```console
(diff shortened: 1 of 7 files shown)


dbrower:
I guess I have no idea which web server curate is using these days. I though it was using unicorn.

msuhovec:
Curate still uses unicorn

PR #790, opened 2019-05-08
Patching critical vulnerability

jeremyf:
I do not believe that Curate exposed this vulnerability, however this
patch is the recommended "monkey patch" if a version upgrade is not
feasible at this time.


jeremyf:
@ialford Here is the monkey patch for the vulnerability

PR #791, opened 2019-05-09
Use token to authenticate API request

laritakr:
Fix logic so API token will mimic token's user to grant access to items.

### DIFF
--- a/app/controllers/api/items_controller.rb
+++ b/app/controllers/api/items_controller.rb
@@ -1,11 +1,17 @@
-class Api::ItemsController < CommonObjectsController
+class Api::ItemsController < ApplicationController
+  respond_to :jsonld
+  include Sufia::Noid # for normalize_identifier method
+  prepend_before_filter :normalize_identifier
+  before_filter :validate_permissions!
+  before_filter :item, only: [:show]
+
   # GET /api/items
   def index
   end
 
   # GET /api/items/1
   def show
-    render json: curation_concern.as_jsonld
+    render json: item.as_jsonld
   end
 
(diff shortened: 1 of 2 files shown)


app/models/curate/user_behavior.rb
```diff
@@ -30,6 +30,7 @@ module UserBehavior
     # Format the json for select2 which requires just an id and a field called text.
     # If we need an alternate format we should probably look at a json template gem
     def as_json(opts = nil)
+      require 'byebug'; debugger; true
```
dbrower: Is this line committed on purpose?

app/controllers/api/items_controller.rb
```diff
@@ -51,4 +56,14 @@ def set_current_user!
...
     end
+
+    def validate_permissions!
+      set_current_user!
+      item_id = params[:id]
+      user_name = @current_user.try(:username) || @current_user
+      # if unauthorized
+      respond_with item do |format|
+        format.jsonld { render json: { error: 'Unauthorized', user: user_name, item: item_id }, status: 401 }
+      end unless current_ability.can? :read, item_id
```
dbrower: one nit....the multiline block with a trailing `unless...` clause is confusing. Can the conditional be pulled to the top?

app/models/curate/user_behavior.rb
laritakr: fixed and force-pushed

app/controllers/api/items_controller.rb
laritakr: changed/simplified this along with added specs

PR #792, opened 2019-05-09
Curate API

laritakr:
Initial Curate API work, including API access token management
Supports  index with some basic searches, show, and file download (includes DLTP-1666 and DLTP-1653)


app/controllers/api/access_tokens_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    # GET /api/access_tokens
+    def index
+      @api_access_tokens = api_access_token_list
+    end
+
+    # GET /api/access_tokens/new
+    def new
+      @api_access_token = ApiAccessToken.new(new_api_access_token_params)
+      create if params.has_key? :api_access_token
```
jeremyf: This line is a bit surprising. I would not expect a `GET` request to update the database. Could you help me understand the purpose of this line.

app/controllers/api/access_tokens_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    # POST /api/access_tokens
+    def create
+      @api_access_token = ApiAccessToken.new(api_access_token_params_with_user)
+      if !@api_access_token.user_id.nil? && @api_access_token.save
```
jeremyf: I would expect the `@api_access_token.save` method to check that we have a `user_id`

Can we add `validates_presence_of :user_id` to the `ApiAccessToken` class? If not, would `@api_access_token.user_id.present?` be equivalent to `!@api_access_token.user_id.nil?`? I find the former easier to quickly read.

app/controllers/api/access_tokens_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+        @api_access_token = ApiAccessToken.find(params[:id])
+      end
+
+      def api_access_token_params
+        params.require(:api_access_token).permit(:user_id, :issued_by)
+      end
+
+      def new_api_access_token_params
+        if params.has_key? :api_access_token
+          params.require(:api_access_token).permit(:user_id, :issued_by)
```
jeremyf: Can you call the `api_access_token_params` method to encapsulate the parameter require/permit behavior?

app/controllers/api/access_tokens_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+        else
+          {}
+        end
+      end
+
+      def api_access_token_params_with_user
+        param_list = api_access_token_params.merge!({ :issued_by => current_user.user_key })
+        if cannot? :manage, ApiAccessToken
+          param_list[:user_id] = params[:api_access_token][:user_id]
```
jeremyf: This will fail if `params[:api_access_token]` is `nil`, which could happen based on my reading of the `new_api_access_token_params`

app/controllers/api/items_controller.rb
```diff
@@ -0,0 +1,216 @@
...
+      render json: { error: 'Item not found', user: user_name, item: item_id }, status: :not_found
+    end
+
+    def build_api_query(solr_parameters, user_parameters)
+      # translate API query terms into solr query
+      Api::QueryBuilder.new(@current_user).build_filter_queries(solr_parameters, user_parameters)
+    end
+end
+
+class Api::QueryBuilder
```
jeremyf: I like this encapsulating class!

app/controllers/common_objects_controller.rb
```diff
@@ -26,6 +26,21 @@ def unauthorized_template
     end
   end
 
+  # Overriding this from Hydra::AccessControlsEnforcement.
+  # The code is identical but moving it seemed to solve a caching issue with Hydra::PermissionsCache,
+  # causing a user to not have permission to view their newly created work,
+  # and requiring a restart of the rails server to reset it.
+  # May need to override in catalog_controller as well, but only doing it here for now.
+  def enforce_show_permissions(opts={})
```
jeremyf: Great catch. And thanks for the documentation!

app/presenters/api/items_search_presenter.rb
```diff
@@ -0,0 +1,224 @@
...
+        "itemUrl" => File.join(request_url.root.to_s,  Rails.application.routes.url_helpers.api_item_path(item_id))
+      }
+      # always include any fields which were part of query.
+      results_hash = results_hash.merge(load_query_fields)
+      # return any additional requested fields
+      if !fields.nil?
+        fields.each do |field|
+          begin
+            results_hash[field.camelize(:lower)] = self.send("include_#{field}")
+          rescue NoMethodError
```
jeremyf: I'm wondering if the following structure might help in future debugging:

```ruby
fields.each do |field|
  next unless respond_to("include_#{field}")
  results_hash[field.camelize(:lower)] =send("include_#{field}")
end
```

This prevents the loop from swallowing all NoMethodError's that may bubble up in the called method.

app/presenters/api/items_search_presenter.rb
```diff
@@ -0,0 +1,224 @@
...
+      results_hash = {
+        "id" => item_id,
+        "title" => dc_title,
+        "type" => dc_type,
+        "itemUrl" => File.join(request_url.root.to_s,  Rails.application.routes.url_helpers.api_item_path(item_id))
+      }
+      # always include any fields which were part of query.
+      results_hash = results_hash.merge(load_query_fields)
+      # return any additional requested fields
+      if !fields.nil?
```
jeremyf: Below I see `if fields.present?`, could that be used? I mentally stumble when reading "if not nil?"

app/presenters/api/show_item_presenter.rb
```diff
@@ -0,0 +1,289 @@
...
+  def process_datastreams(object)
+    data = {}
+    object.datastreams.each do |dsname, ds|
+      next if dsname == 'DC'
+      method_key = "process_#{dsname.gsub('-', '')}".to_sym
+      if respond_to?(method_key, true)
+        parsed_data = self.send(method_key, ds)
+        data = merge_hashes(data, parsed_data)
+      else
+        Rails.logger.error("#{item_id}: unknown datastream #{dsname}")
```
jeremyf: I'm wondering if we should also send this to Sentry, as its a bit of a surprise. But not something to throw an exception.

app/views/api/items/show.html.erb
```diff
@@ -0,0 +1 @@
+<p>STUB</p>
```
jeremyf: A bit curious about this.

app/controllers/api/access_tokens_controller.rb
laritakr: Only admin users see the form to select a user for the API access token being created. Other users can only create access tokens for themselves, so the new request passes in the info and goes directly to the create option.

app/controllers/api/access_tokens_controller.rb
laritakr: I believe that should work. I'll double-check. I always forget about using the built-in validates methods.

app/presenters/api/items_search_presenter.rb
laritakr: That should be the same. I will try it and make sure. I struggle with nil vs blank vs present vs empty. I think I may have tried empty instead of present, and ended up using nil.

app/presenters/api/show_item_presenter.rb
laritakr: Probably a good idea. Since I started with the existing json parser, I just mimicked the logic there (in `as_jsonld_mapper.rb`) and didn't go back to add in Sentry notification.

app/views/api/items/show.html.erb
laritakr: I will remove this again. It was part of Mark's initial work, and once I coded the methods, I deleted it. It apparently reappeared during my mess up with the branch  last week and I didn't notice.

app/controllers/api/access_tokens_controller.rb
laritakr: The form always assigns a user into `params[:api_access_token][:user_id]`, or is bypassed if we already have a user. The user_id is always present when we get to this point.

PR #794, opened 2019-05-15
Add search to API

laritakr:
### Supports text search params:
- q=basic+text&search_field=all_fields

### Supports additional filtering by:
- type= model or document subtype
- editor= edit access person
- creator= depositor

### Includes standard pagination options:
- page= specify current page
- rows= specify items per page

### Flexible results set
- Always shows pid, title, type.
- Always shows fields included in filtered options.
- Includes option to specify additional fields to show:
-depositor
-creator
-language

### Sample search results
Screen Shot 2019-05-15 at 1 56 28 PM


app/presenters/api/items_search_presenter.rb
```diff
@@ -0,0 +1,209 @@
...
+      @item.fetch('id')
+    end
+
+    def dc_title
+      dc_title = Array.wrap(@item.fetch('desc_metadata__title_tesim', 'title-not-found' )).first
+    end
+
+    def dc_type
+      model = Array.wrap(@item.fetch('active_fedora_model_ssi'))
+      dc_type = @item.fetch('desc_metadata__type_tesim', model)
```
jeremyf: Do we need to wrap dc_type in an array as well? I suspect we could do the following:

```ruby
model = @item.fetch('active_fedora_model_ssi')
dc_type = @item.fetch('desc_metadata__type_tesim', model)
Array.wrap(dc_type).first
```

laritakr: I'm not sure but I did to be safe.

PR #802, opened 2019-05-23
Format API Show Results

laritakr:
Creates new json format for api/items/xxxxxx response.

### DIFF
--- a/app/services/merge_hash.rb
+++ b/app/services/merge_hash.rb
@@ -1,0 +1,34 @@
+class MergeHash
+  # we need to merge the hashes, such that we
+  # 1) combine data with the same keys
+  # 2) do not duplicate any values which are in both
+  # 3) do not lose keys from either hash.
+  def merge_hashes(h1, h2)
+    # first, we do a standard merge
+    combined_hash = h2.merge(h1)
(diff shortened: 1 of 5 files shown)


app/services/merge_hash.rb
```diff
@@ -0,0 +1,34 @@
+class MergeHash
+  # we need to merge the hashes, such that we
+  # 1) combine data with the same keys
+  # 2) do not duplicate any values which are in both
+  # 3) do not lose keys from either hash.
+  def merge_hashes(h1, h2)
+    # first, we do a standard merge
+    combined_hash = h2.merge(h1)
+    # then we do extra processing, with hashes removed, to add in anything which would have gotten excluded by the first merge.
+    h1.merge(h2) do |key, x1, x2|
```
dbrower: I'm curious why there is a need to calculate `combined_hash` in line 8. I would have thought  doing a merge with a block to handle duplicate keys would be enough.

laritakr: The block to handle duplicate keys caused a loss of keys that were in the first hash and not in the second. I had to do it both ways for it to be complete.

PR #808, opened 2019-06-18
Curate authority management

laritakr:
Use Authority Groups to authorize functions in Curate.

- Adds new DB table migrations.
- Adds UI to add and manage Admin::AuthorityGroups, and link them to Hydramata::Groups.
- Adds control to disallow deleting Hydramata::Groups which are used in Admin::AuthorityGroups.
- Uses current admin yml to initialize super admin AuthorityGroups with current RepositoryAdministrator permissions.
- Hides all options on the Repository Administration menu from `admin` group other than the option to escalate privileges. Only `super_admin` can see the other options.

Initially creates 4 authority groups and associated classes:
- SuperAdmin (super_admin) ~> control site authority via UI, view full Administrative menu, along with all RepositoryAdministrator abilities.
- RepositoryAdministrator (admin) ~> ability to escalate authority to edit anything site-wide
- TokenManager (token_manager) ~> ability to clean up expired time-limited tokens
- ViewAll (view_all) ~> ability to view all works site-wide, regardless of work permissions. No edit rights on works unless also have admin rights.

Also fixes an issue with masquerade routing, where masquerades did not reset to original user, because masquerades are very helpful for testing authority via the UI!

References https://jira.library.nd.edu/browse/DLTP-1569


app/controllers/admin/authority_groups_controller.rb
```diff
@@ -0,0 +1,88 @@
...
+  def show
+  end
+
+  def create
+   @admin_authority_group = Admin::AuthorityGroup.new(admin_authority_group_params)
+   if @admin_authority_group.save
+     @notice = 'Authority Group was successfully created.'
+     redirect_to admin_authority_groups_path, notice: @notice
+    else
+      @notice = 'Authority Group was not created.'
```
jeremyf: I suspect you'll need a `flash.now` similar to the `#update` method.

app/models/admin/authority_group.rb
```diff
@@ -0,0 +1,52 @@
...
+      return [] if self.authority_group(auth_group_name: auth_group_name).nil?
+      return [] if @authority_group.associated_group.nil?
+      @members = []
+      @authority_group.associated_group.members.each do |person|
+        @members << person.name
+      end
+      @members
+    end
+
+    def self.authority_group(auth_group_name:)
```
jeremyf: Don't use a class instance variable, use a local variable instead. The class instance variable will persist between calls to this method and will sit in memory of the application.

```ruby
def self.authority_group(auth_group_name:)
  begin
    authority_group = AuthorityGroup.find_by(auth_group_name: auth_group_name)
  rescue
    return nil
  end
  authority_group
end
```

db/schema.rb
```diff
@@ -384,6 +397,7 @@
     t.datetime "updated_at"
   end
 
+  Foreigner.load
```
jeremyf: I'm not certain what this is about.

laritakr: I have no idea where that came from. Did I do something that could have caused it?

laritakr: Based on google, this apparently comes from the foreigner gem, which is part of mailboxer. This may be the first time we updated the schema since the rails version update?  https://github.com/matthuhiggins/foreigner/issues/161

PR #809, opened 2019-07-01
Add perf gems

msuhovec:


Gemfile
```diff
@@ -52,7 +53,7 @@ group :default do
   # Need rubyracer to run integration tests.....really?!?
   # See https://github.com/sstephenson/execjs#readme for more supported runtimes
   gem 'libv8', '~> 3.16.14.3'
-  gem 'locabulary', github: 'ndlib/locabulary', ref: 'b8ab510dce637d37229d001fedfbc6af1ab510f2'
+  gem 'locabulary', github: 'ndlib/locabulary', ref: '9d4a7c1b6a5956c5152924b1ff51d3f0798581d8'
```
dbrower: This is reverting to the previous version of locabulary? (see commit 6f23df574e838472433a622344765b231b470da0)

dbrower: and also commit 458b7c9fffd0f68f08cd78e8ef624a62e1059c8f.

dbrower:
Why are the stop words in this commit? Those files don't even look changed.

I see that you noticed and fixed the locabulary version change. Can you squash these two commits and push the branch back up?

Gemfile
msuhovec: went to lobabulary b8ab510dce637d37229d001fedfbc6af1ab510f2

PR #834, opened 2019-08-22
Curate api trx append

msuhovec:


app/controllers/api/uploads_controller.rb
```diff
@@ -37,39 +45,54 @@ def trx_new_file
       content = s3.bucket(ENV['S3_BUCKET']).object("#{trx_id}/#{file_pid}-#{next_sequence}")
       #copy body of message to bucket
       content.put(body: request.body())
-      metadata = s3.bucket(ENV['S3_BUCKET']).object("#{trx_id}/metadata-#{file_pid}.json")
-      metadata.put(body: initial_file_metadata(file_name, file_pid))
+      metadata = s3.bucket(ENV['S3_BUCKET']).object("#{trx_id}/metadata-file-#{file_pid}.json")
```
jeremyf: I see duplication in line 48 and line 17. Would it be feasible to extract a method to isolate some of that logic?

My personal foible says "Keep as much logic out of the controller" but I'm not prepared to die on that hill. I'm wondering if it makes sense to extract an object that handles much of this logic?

laritakr: We are trying to just get things working and then encapsulate the duplication. I'd prefer to leave it as is for now.

laritakr: But yes, long-term, the goal is to extract an object.

msuhovec: Larita and I have gone back and forth on  the controller vs app/model thing (well, maybe app/services in this case). I think we will end up doing that.

msuhovec: I have talked about consolidating the S3 stuff- I will do that now.

jeremyf: Excellent. Looks functional and I'm glad to hear that you're having conversations about approaches.

PR #837, opened 2019-09-03
Link to ArchivesSpace

laritakr:
Attempts to view finding aids which are private will show a link to archivesspace rather than the unauthorized page.

https://jira.library.nd.edu/browse/DLTP-1711

Screen Shot 2019-09-03 at 3 28 11 PM

### DIFF
--- a/app/views/curation_concern/base/unauthorized.html.erb
+++ b/app/views/curation_concern/base/unauthorized.html.erb
@@ -1,8 +1,16 @@
+<% work_type = curation_concern.human_readable_type %>
+
+<% if work_type == 'Finding Aid' %>
+<h1>Content Moved</h1>
+  <p>This item has been moved to <a href="https://archivesspace.library.nd.edu">archivesspace.library.nd.edu.</p>
+
+<% else %>
 <h1>Unauthorized</h1>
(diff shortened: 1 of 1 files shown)


app/views/curation_concern/base/unauthorized.html.erb
```diff
@@ -1,8 +1,16 @@
+<% work_type = curation_concern.human_readable_type %>
+
+<% if work_type == 'Finding Aid' %>
+<h1>Content Moved</h1>
+  <p>This item has been moved to <a href="https://archivesspace.library.nd.edu">archivesspace.library.nd.edu.</p>
```
dbrower: </a>

laritakr: oops! Thanks for catching that.

Also wondering the best way for checking for finding aid types? Should I do something like `curation_concern.is_a?(FindingAid)`? @jeremyf

jeremyf: `.is_a?(FindingAid)` would be reasonable. Unless we had some kind of "canonical URL" attribute on the given object.

laritakr: ok both are fixed.

PR #838, opened 2019-09-04
Process API Callbacks

laritakr:
Authenticates callback and updates transaction with status from batch ingest.

Also fix storage bucket logic. It worked for specs but not for other environments.

### DIFF
--- a/app/controllers/admin/callbacks_controller.rb
+++ b/app/controllers/admin/callbacks_controller.rb
@@ -13,12 +13,12 @@ module Admin
     # params: {"controller"=>"admin/callbacks", "action"=>"callback_response", "tid"=>"c7af186a8d75c44a", "response"=>"ingest_completed", "format"=>"json"}
 
     def callback_response
+      @trx_id = params[:tid]
       response = params[:response]
 
       if validated
         response_data = JSON.parse(request.body.read)
         update_transaction_status_based_on(trx_id: trx_id, ingest_status: response_data["job_state"])
-
         render json: { trx_id: trx_id, ingest_response: response_data["job_state"] }, status: :ok
       else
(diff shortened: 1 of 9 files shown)


app/controllers/admin/callbacks_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+    private
+
+    def validate_authentication
+      @validated = authenticate_with_http_basic do |user, pass|
+        authenticate_trx(user_name_hash: user, trx_id_hash: pass)
+      end
+      @validated
+    end
+
+    def update_transaction_status_based_on(trx_id:, ingest_status:)
```
jeremyf: This method feels odd. In part because the name `.set_status` feels as though it should change state. When it instead translates a status. In addition, the knowledge of state transitions for an ApiTransaction are not contained within the ApiTransaction. I assume there are other points in the code that update to different transaction states based on a different status.

I'm thinking of a Finite Statemachine and how it is often implemented in Ruby (https://www.sitepoint.com/state-machines-in-ruby/).

```ruby
class ApiTransaction
  def self.update_based_on_ingest_status(trx_id:, ingest_status:)
    # Contain the logic
  end
end
```

app/controllers/admin/callbacks_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+      when "error"
+        ApiTransaction.update(trx_id, trx_status: ApiTransaction.set_status(:error))
+      when "processing"
+        # do nothing
+      else
+        # do nothing
+      end
+    end
+
+    def authenticate_trx(user_name_hash:, trx_id_hash:)
```
jeremyf: Would this be easier to test in a service object? There are a lot of conditionals, and to test each one seems to require a lot of effort.

PR #847, opened 2019-10-16
Fix failing derivative generation

laritakr:
Remove `transform_datastream` blocks. Derivatives cannot be generated with
two blocks... the second one fails. 

Also remove the option to upload an account image from the edit view. This option was not working and no one can see the images anyway, since person was removed from the catalog search results list.

### DIFF
--- a/app/repository_models/collection.rb
+++ b/app/repository_models/collection.rb
@@ -56,7 +56,7 @@ class Collection < ActiveFedora::Base
   def generate_derivatives
     case mime_type
     when 'image/png', 'image/jpeg', 'image/tiff'
-      transform_datastream :content, { medium: {size: "300x300>", datastream: 'medium'}, thumb: {size: "100x100>", datastream: 'thumbnail'} }
+      transform_datastream :content, { medium: {size: "300x300", datastream: 'medium'} }
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/repository_models/collection.rb
```diff
@@ -56,7 +56,7 @@ def representative
   def generate_derivatives
     case mime_type
     when 'image/png', 'image/jpeg', 'image/tiff'
-      transform_datastream :content, { medium: {size: "300x300>", datastream: 'medium'}, thumb: {size: "100x100>", datastream: 'thumbnail'} }
+      transform_datastream :content, { medium: {size: "300x300>", datastream: 'medium'} }
```
dbrower: could you remove the medium one and keep the thumbnail one instead? I dont think we *ever* return the `medium` datastream, and having something in the thumbnail datastream is useful for the file display. (e.g. https://curate.nd.edu/show/und:xg94hm53g8t)

laritakr: Yes I'm changing all medium to thumbnail. But the video didn't seem to create a thumbnail, and this change didn't allow me to update my account profile pic, so I am doing more testing and changes as well.

laritakr: Removing medium caused some issues so I am removing thumbnails leaving that one.

PR #848, opened 2019-10-17
Okta authentication

laritakr:
Began with the `updating to rails 4.2` branch to authenticate with Okta.


dbrower:
I noticed this is a "Draft" PR. How cool.

There are so many commits, and most are of partial work and WIP. Could you squash everything into one commit (or a few, if it makes logical sense) before merging?

laritakr:
> I noticed this is a "Draft" PR. How cool.

Yeah, that's a fairly new feature. This obviously isn't ready to merge so that prevents anyone merging it. :) 

And yes, it makes sense to squash these somewhat, but I might let @jeremyf decide how much squashing to do! This is encompassing both the rails updates and the okta stuff.

PR #852, opened 2019-11-01
Add Usage page

laritakr:
Add metrics database into Curate.
Add classes to access metrics tables.
Add new page to show usage metrics for a work.
Link to new page from Common Objects show page (for viewers only)

### DIFF
--- a/app/controllers/metrics/usage_controller.rb
+++ b/app/controllers/metrics/usage_controller.rb
@@ -1,0 +1,103 @@
+require 'csv'
+
+module Metrics
+  class UsageController < CommonObjectsController
+    respond_to(:html, :csv)
+    attr_reader :curation_concern, :usage_details
+
+    # GET /usage/:id
+    def show
+      noid = Sufia::Noid.noidify(params[:id])
+      if curation_concern.nil?
+        redirect_to root_url, notice: "Unable to display usage. Item #{noid} not found."
+      else
+        respond_to do |format|
+          format.html { parsed_usage_details(noid: noid) }
+          format.csv { send_data format_for_csv(noid: noid), filename: "#{noid}-#{Time.now}.csv" }
+        end
+      end
+    end
+
+    private
+
+    # note: Metrics uses noid as the item pid
+    def usage_query(pid:)
+      @usage_query ||= FedoraAccessEvent.all_usage_including_child_objects_for(pid: pid)
+    end
+
+    def format_for_csv(noid:)
+      usage_query(pid: noid).to_csv
+    end
+
+    def parsed_usage_details(noid:)
+      @usage_details ||= begin
+        parse_usage(usage: usage_query(pid: noid))
+      end
+    end
+
+    def curation_concern
+      @curation_concern ||= begin
+        ActiveFedora::Base.load_instance_from_solr(params[:id])
+      rescue ActiveFedora::ObjectNotFoundError => e
+        Raven.capture_exception(e)
+        nil
+      end
+    end
+
+    # @param usage [Array<FedoraAccessEvent>] an array of Metrics::FedoraAccessEvent
+    # @return [Array<UsageItem>] an array of Metrics::UsageItems
+    def parse_usage(usage:)
+      usage_array = []
+      pid_list = usage.all.group(:pid).pluck(:pid)
+      pid_list.each do |pid|
+        item_events = usage.where(pid: pid)
+        usage_array << UsageItem.new(noid: pid, metrics: item_events) if item_events.any?
+      end
+      usage_array
+    end
+  end
+
+  # A class to organize the events and usage counts for one pid
+  class UsageItem
+    attr_reader :noid, :label, :events
+
+    # @param metrics [Metrics::UsageEvent]
+    def initialize(noid:, metrics:)
+      @noid = noid # String
+      @label = item_label(noid: noid) # String
+      @events = item_events(events: metrics) # Array<Metrics::ItemUsageEvent>
+    end
+
+    # @param noid [String] noid of item
+    # @return [String] Title of item
+    def item_label(noid:)
+      id = Sufia::Noid.namespaceize(noid)
+      begin
+        item = ActiveFedora::Base.load_instance_from_solr(id)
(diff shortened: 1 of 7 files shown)


laritakr:
Dependent on okta deployment work to be completed, and the addition of the new keys as added in application.yml.

laritakr:
Sample screenshot of the new page: 
Screen Shot 2019-11-01 at 2 07 57 PM

dbrower:
Looks good! I have some text edits I would like (@mikalarae, thoughts?), maybe you could make them before merging this?
* Change past year to "past 365 days"
* Change all time to "Since Deposit"
* Add following text to top of page: "Usage information for this item and any attached files."
* Add following text to bottom of page:

A _view_ is any display of the information page for this item in CurateND. A _download_ is the number of times either this file has been downloaded (for files) or the total number of times any file belonging to this item has been downloaded (for works).

Collection of usage data started June 2016. To get the complete usage data for this item going back to June 2016, use the "Export to CSV" option.

laritakr:
@don The metrics are actually since we started calculating metrics... not necessarily since deposit. Is that a problem?

There is no export to csv yet. I assume that's what you want me to work on with this next? (along with some specs).

If we want to change the view to be accessible to those with only view rights, we could either change change the object links to just be text, or check permissions on each individual object before including it at all.

app/controllers/metrics/usage_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+  class UsageController < ApplicationController
+    with_themed_layout '1_column'
+    respond_to(:html)
+    include Sufia::Noid # for normalize_identifier method
+    prepend_before_filter :normalize_identifier
+    attr_reader :curation_concern, :usage_details
+
+    # GET /usage/:id
+    def show
+      @id = params[:id]
```
jeremyf: Is it necessary to have an instance_variable for `@id`; It looks as though you are interested in `noid` and nothing more.

(I'm always looking at ways to reduce the number of instance variables declared within a controller method).

app/controllers/metrics/usage_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+      # note: Metrics uses noid in column :pid
+      noid = Sufia::Noid.noidify(@id)
+      usage_query = FedoraAccessEvent.all_usage_including_child_objects_for(pid: noid)
+      @usage_details = parse_usage(usage: usage_query)
+    end
+
+    def curation_concern
+      @curation_concern ||= begin
+        ActiveFedora::Base.load_instance_from_solr(params[:id])
+      rescue ActiveFedora::ObjectNotFoundError
```
jeremyf: Do we want to quietly suppress this?

app/controllers/metrics/usage_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+    end
+
+    # @param noid [String] noid of item
+    # @return [String] Title of item
+    def item_label(noid:)
+      id = Sufia::Noid.namespaceize(noid)
+      begin
+        item = ActiveFedora::Base.load_instance_from_solr(id)
+        return item.human_readable_type + ": " + item.title
+      rescue
```
jeremyf: What specific are you rescuing? My preference is to identify that in the rescue clause.

app/controllers/metrics/usage_controller.rb
```diff
@@ -0,0 +1,108 @@
...
+        return "Item: " + noid
+      end
+    end
+
+    # @param events [#<ActiveRecord::Relation] Result list of Metrics::FedoraAccessEvent
+    # @return [Array #<UsageItemEvent]
+    def item_events(events:)
```
jeremyf: I'm a bit confused about what `events` is for the parameter. Please document that as well as the purpose of this method; I think there may be an improvement we can make in the computation.

app/models/metrics.rb
```diff
@@ -0,0 +1,57 @@
...
+    # t.datetime "created_at",             null: false
+    # t.datetime "updated_at",             null: false
+
+    def self.all_usage_including_child_objects_for(pid:)
+      usage_events = []
+      pid_list = FedoraObject.ids_of_object_and_all_child_objects_of(pid: pid)
+      if !pid_list.empty?
+        # create an array of events for each pid
+        pid_list.each do |pid|
+          usage_events << { pid:pid,
```
jeremyf: I see a hash, which is used elsewhere. It may be helpful to make a simple class.
```ruby
# Declare outside of the method
UsageEvent = Struct.new(:pid, :events)

# Within method declaration
usage_events << UsageEvent.new(pid, FedoraAccessEvent.where(pid: pid))
```

app/controllers/metrics/usage_controller.rb
laritakr: I suppressed it because I was using the live database connection in a test environment to pull up pages for items that weren't in my AF instance. In real life, I assume we would want to send it to Sentry?

app/controllers/metrics/usage_controller.rb
laritakr: If we don't find the item, I was showing just the id rather than the title (same reason as above). Do we want each of these reported in Sentry?

app/controllers/metrics/usage_controller.rb
laritakr: Perhaps not any more. I was using it but I think I eliminated that use.

app/controllers/metrics/usage_controller.rb
laritakr: I'm don't follow what you don't understand... events is the result of the sql, so it's an ActiveRecord::Relation which is a list of FedoraAccessEvents. How should that be documented differently?

app/controllers/metrics/usage_controller.rb
laritakr: Oops I missed that I didn't specify the rescued error so I misunderstood your question.

laritakr:
Screen Shot 2019-11-08 at 3 19 55 PM

laritakr:
Closing in order to create a PR on master instead.

PR #866, opened 2019-12-10
Remove PlumX and add Usage view text

laritakr:
DLTP-1720 & DLTP-1733


dbrower:
can you also remove the code block calculating `has_doi`, since that isn't needed either?

laritakr:
oops. sure!

PR #876, opened 2020-01-29
Create architecture dissertation work type

laritakr:
Create new work type, based on ETD, with addition of a dissertation type. 

Type is metadata content only and does not behave as type in document... i.e. it does not change the metadata fields for the work subtype. It does get used as work type for displays and faceting. 

https://jira.library.nd.edu/browse/DLTP-1748


config/work_type_policy_rules.yml
```diff
@@ -8,10 +8,12 @@ CatholicDocument:
   open: all
 Dataset:
   open: all
+Dissertation:
+  open: all
 Document:
   open: all
 Etd:
-  open: nobody
+  open: all
```
jeremyf: Can you tell me more about this change?

laritakr: This only affects running it locally, and I changed it so I could compare ETD to my new dissertation type. In production, it is controlled by a group id. I didn't change anything in any of the staging or production environments.

jeremyf: Cool. That sounds like a great change.

app/assets/stylesheets/curate.scss
```diff
@@ -28,6 +28,6 @@
 @import 'style/theme';
 @import 'style/typography';
 
-.negetive-left-margin {
+.negative-left-margin {
```
jeremyf: Good catch on spelling!

PR #878, opened 2020-01-29
Remove crosshatch background

laritakr:


dbrower:
looks fine. Should we also remove the `bg.png` file?

laritakr:
That's a blacklight file in the vendor directory, so i'd leave it.

PR #884, opened 2020-02-24
Correctly identify a bendo url

laritakr:
Bendo url logic was too fragile. Logic was changed to not validate the url as bendo... if a url exists with a redirect, we are assuming it is bendo.

### DIFF
--- a/app/services/bendo/datastream_presenter.rb
+++ b/app/services/bendo/datastream_presenter.rb
@@ -4,7 +4,7 @@ module Bendo
   class DatastreamPresenter
     attr_reader :datastream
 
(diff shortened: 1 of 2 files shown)


app/services/bendo/datastream_presenter.rb
```diff
@@ -6,7 +6,6 @@ class DatastreamPresenter
 
     def initialize(datastream:, service_url: Bendo.url)
```
dbrower: someone is probably calling this method? can they not pass in `service_url`? Oh wait! that's a default parameter? can it be removed?

jeremyf:
:)

PR #886, opened 2020-02-25
Add visibility badge label

laritakr:
Add a label next to each visibility badge on the catalog results page.

### DIFF
--- a/app/views/catalog/_document.html.erb
+++ b/app/views/catalog/_document.html.erb
@@ -33,7 +33,8 @@
           <dt>Date Created: </dt>
           <dd><%= document.created %></dd>
         <% end %>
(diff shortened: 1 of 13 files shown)


app/services/bendo/datastream_presenter.rb
```diff
@@ -4,9 +4,8 @@ module Bendo
   class DatastreamPresenter
     attr_reader :datastream
 
-    def initialize(datastream:, service_url: Bendo.url)
+    def initialize(datastream:)
```
dbrower: This looks like you are pulling in changes from #884 ?

laritakr: Only because that's not merged yet. I built on that one, that so that once that's merged this won't have to build again.

PR #887, opened 2020-02-27
Dltp 1783

msuhovec:
fixing return code login in trx_commit

### DIFF
--- a/app/controllers/api/uploads_controller.rb
+++ b/app/controllers/api/uploads_controller.rb
@@ -127,7 +127,7 @@ class Api::UploadsController < Api::BaseController
       # update trx status
       ApiTransaction.set_status_based_on(trx_id: trx_id, action: :commit)
       # submit to ingestor
-      ingestor_status = BatchIngestor.start_api_ingest(trx_id).code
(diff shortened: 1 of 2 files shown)


app/controllers/api/uploads_controller.rb
```diff
@@ -127,12 +127,8 @@ def trx_commit
       # update trx status
       ApiTransaction.set_status_based_on(trx_id: trx_id, action: :commit)
       # submit to ingestor
-      ingestor_status = BatchIngestor.start_api_ingest(trx_id).code
-      if ingestor_status == '200'
-        render json: { trx_id: trx_id }, status: :ok
-      else
-        render json: { trx_id: trx_id, error: 'Error submitting ingest request' }, status: :bad_request
-      end
+      BatchIngestor.start_api_ingest(trx_id)
```
jeremyf: It looks like this removed an error check status. Is that expected?

msuhovec:
It was originally coded with the expectation that the underlying Batch
Ingester class did not handle http error code itself, but merely reported
them. This is not the case- the Batch ingester class does error handling
and logging of http errors, and tells the caller that the Ingest
succeeded/failed.

PR #895, opened 2020-03-27
Dltp 1758 oai

laritakr:
Completes a portion of https://jira.library.nd.edu/browse/DLTP-1758
 
ListRecords verb is still unsupported, waiting for some changes in the oai gem. 
Additional fields can easily be added to the response by updating the field list in `app/models/oai_provider/curate_format.rb`


app/models/oai_provider/curate_format.rb
```diff
@@ -0,0 +1,27 @@
+class OaiProvider < OAI::Provider::Base
```
jeremyf: How would you feel about moving the CurateFormat class and CurationConcernProvider class into the `app/models/oai_provider.rb` I don't see a strong reason to keep these files separate.

laritakr: I would be fine with that. I started out thinking more files would be needed but it ended up being pretty minimal.

jeremyf: Feel free to consolidate, I'll continue my review of what you have.

jeremyf:
Looks good, I'll "sign off" after the consolidation.

app/models/oai_provider/curate_format.rb
laritakr: Done

PR #897, opened 2020-04-14
OAIPMH query search

laritakr:
1. Support ListIdentifiers and ListRecords verbs (for from & until searches), along with resumptionToken to paginate through records.
2. Support ListSets
3. Support ListRecords search for set (collection & model searches)
* collection:und:xxxxxxxx selects subcollection and work members of a LibraryCollection
* model:Article allows search any registered CurationConcern and returns paginated list of items of that model

Completes a portion of https://jira.library.nd.edu/browse/DLTP-1758

Does not yet include specs

### DIFF
--- a/app/controllers/oai_controller.rb
+++ b/app/controllers/oai_controller.rb
@@ -16,10 +16,6 @@ class OaiController < CatalogController
     render body: response, content_type: "text/xml"
   end
 
-  def valid_search_request_syntax?(options)
-    Oai::QueryBuilder.new.valid_request?(options)
-  end
-
   private
 
     def oai_params
--- a/app/controllers/oai_controller.rb
+++ b/app/controllers/oai_controller.rb
@@ -16,24 +16,14 @@ class OaiController < CatalogController
     render body: response, content_type: "text/xml"
   end
 
-  def load_data(options = {})
-    if valid_search_request_syntax?(options)
-      params.merge!(options)
-      # returns [ Blacklight::SolrResponse, Array(SolrDocument) ]
-      (@response, @document_list) = get_search_results
-    else
-      return nil
-    end
+  def valid_search_request_syntax?(options)
+    Oai::QueryBuilder.new.valid_request?(options)
   end
 
   private
 
     def oai_params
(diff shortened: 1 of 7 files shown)


app/controllers/oai_controller.rb
```diff
@@ -1,20 +1,42 @@
...
+  ]
+
   def index
-    provider = OaiProvider.new
+    provider = OaiProvider.new(controller: self)
     response = provider.process_request(oai_params.to_h)
     render body: response, content_type: "text/xml"
   end
 
+  def load_data(options = {})
```
jeremyf, line 19: Can this method be added to the CurationConcern provider? I'm wondering how important it is to set an instance variable on the controller. My understanding is the CurationConcern provider needs the returned value of get_search_results.

To test if it can be extracted, I'm wondering how things would work if you were to turn the `@response` and `@document_list` into local variables (e.g. `response`, `document_list`)

app/controllers/oai_controller.rb
```diff
@@ -1,20 +1,32 @@
...
+    provider = OaiProvider.new(controller: self)
     response = provider.process_request(oai_params.to_h)
     render body: response, content_type: "text/xml"
   end
 
-  private
-
-  def oai_params
-    params.permit(:verb, :identifier, :metadataPrefix, :set, :from, :until, :resumptionToken)
+  def valid_search_request_syntax?(options)
```
jeremyf, line 19: Can this be moved into the provider? It looks like it can.

app/services/oai/query_builder.rb
```diff
@@ -0,0 +1,126 @@
...
+  }.freeze
+  VALID_SET_TYPE_SEARCHES = {
+    collection: "und:",
+    model: "type"
+  }
+
+  attr_reader :current_user
+
+  def initialize
+    @current_user = nil
```
jeremyf, line 16: Tell me more about this being set to nil?

laritakr: Oops... I set current user to nil because we are only returning public records via OAI and I began this service based on the api query builder. But I removed all of the methods using current_user in here. So it's not actually necessary any more.

PR #903, opened 2020-05-05
Update faqs.html.erb

mikalarae:
Updated FAQs to be more consistent with repository goals.

### DIFF
--- a/app/views/static_pages/faqs.html.erb
+++ b/app/views/static_pages/faqs.html.erb
@@ -138,6 +138,7 @@
         A Creative Commons license is one of a <a href="http://creativecommons.org/licenses/">set of licenses</a> providing others various degrees of permission for use.
         Some of these permissions require attribution, allow for commercial use and re-use, or permit others to make derivative works.
       </dd>
+<dd>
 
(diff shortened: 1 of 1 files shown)


app/views/static_pages/faqs.html.erb
```diff
@@ -57,42 +57,43 @@
     <dl>
       <dt>What is CurateND?</dt>
       <dd>
-        CurateND is the institutional repository for the University of Notre Dame.
+        CurateND is the institutional repository for the University of Notre Dame managed by Hesburgh Libraries.
```
laritakr, line 60: I'd like to see a comma prior to "managed by Hesburgh Libraries"

app/views/static_pages/faqs.html.erb
```diff
@@ -137,38 +138,23 @@
         Some of these permissions require attribution, allow for commercial use and re-use, or permit others to make derivative works.
       </dd>
 
-      <dt>What other licensing options are available?</dt>
```
laritakr, line 140: I'm curious why the licensing options are being removed? Are we putting a list somewhere else?

PR #910, opened 2020-05-21
Remove plumx widget script

dbrower:
Tricky since the script doesn't have plumx in the name. It in turn was
loading other plu.mx resources, which is how I found it.


jeremyf:
I will always merge and approve any commit that removes javascript!

PR #920, opened 2020-06-04
Fix Usage page

laritakr:
Permission check failed when the user wasn't logged in.

Also handle errors when there is no database defined to allow page to
function in development.

### DIFF
--- a/app/controllers/metrics/usage_controller.rb
+++ b/app/controllers/metrics/usage_controller.rb
@@ -47,13 +47,12 @@ module Metrics
     # @param usage [Array<FedoraAccessEvent>] an array of Metrics::FedoraAccessEvent
     # @return [Array<UsageItem>] an array of Metrics::UsageItems
     def parse_usage(usage:)
+      pid_list = usage ? usage.all.group(:pid).pluck(:pid) : []
       usage_array = []
-      pid_list = usage.all.group(:pid).pluck(:pid)
       pid_list.each do |pid|
         item_events = usage.where(pid: pid)
         usage_array << UsageItem.new(noid: pid, metrics: item_events) if item_events.any?
       end
-      rescue NoMethodError
       usage_array
     end
   end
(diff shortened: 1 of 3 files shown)


app/controllers/metrics/usage_controller.rb
```diff
@@ -53,6 +53,7 @@ def parse_usage(usage:)
         item_events = usage.where(pid: pid)
         usage_array << UsageItem.new(noid: pid, metrics: item_events) if item_events.any?
       end
+      rescue NoMethodError
```
jeremyf, line 56: I think this may have some unexpected behavior.

The way this works is now if we encounter a "NoMethodError", we'll return usage_array. Otherwise, we return the results of the above each statement (which will not be the usage_array).

This would require a begin, rescue, end block.

app/models/metrics.rb
```diff
@@ -19,7 +19,11 @@ class FedoraAccessEvent < Metrics::Base
     # @param pid [String] the noid of an ActiveFedora item
     # @return [Metrics::FedoraAccessEvent::ActiveRecord_Relation]
     def self.all_usage_including_child_objects_for(pid:)
-      usage_events = FedoraAccessEvent.where(pid: FedoraObject.ids_of_object_and_all_child_objects_of(pid: pid))
+      begin
+        usage_events = FedoraAccessEvent.where(pid: FedoraObject.ids_of_object_and_all_child_objects_of(pid: pid))
+      rescue Mysql2::Error
```
jeremyf, line 24: Hmm. Do you want a nil object? How will that be handled upstream by the caller of this method?

app/controllers/metrics/usage_controller.rb
laritakr: I decided we probably didn't want to rescue NoMethodError, so I got rid of the error handling and just tested for a nil value instead.

app/models/metrics.rb
laritakr: I replaced the NoMethodError with a direct test for nil.

PR #929, opened 2020-08-10
OAI updates

laritakr:
Resolves CURATE-119

1. Creates a module adding a standardize method to Collections, Works, and Generic Files. This will allow the creation of a common group of metadata terms which can be shared by all models.
2. Rename OaiProvider to CurateOaiProvider and update values shown via Identify verb.
3. Change to standard DC terms for oai_dc metadataPrefix.
4. Add dcterms metadataPrefix.
5. Add new Primo set.


app/services/oai/query_builder.rb
```diff
@@ -25,7 +27,7 @@ def build_filter_queries(solr_parameters, user_parameters)
     user_parameters.each do |term, value|
       (search_key, search_value) = search_terms_for(term, value)
       if VALID_KEYS_AND_SEARCH_FIELDNAMES[search_key].present?
-        solr_parameters[:fq] << do_search_for(search_key, search_value)
+        solr_parameters[:fq] += Array.wrap(do_search_for(search_key, search_value))
```
jeremyf, line 30: Moving from `<<` to `+=` may be creating an issue.

Consider the use of `+=`:

```ruby
>> a = {fq: []}
=> {:fq=>[]}
>> a[:fq].object_id
=> 70309860452960
>> b = a
=> {:fq=>[]}
>> b[:fq] ||= []
=> []
>> b[:fq] += [1,2,3]
=> [1, 2, 3]
>> b[:fq].object_id
=> 70309852445800
>> a[:fq].object_id
=> 70309852445800
```

When we use `b[:fq] += [1,2,3]` the object_id of `b[:fq]` changes.  This is because `+=` is a short-cut for `b = b + [1,2,3]`; Which creates a new array.  What I find odd, and would think it would allow this to work is that `a[:fq].object_id == b[:fq].object_id`.  I’m vaguely aware that later versions of Ruby may have more considerations around this kind of mutation.  _Note: I include `a` and `b` because ruby passes parameters by reference (e.g. the input `solr_parameter` is defined elsewhere and passed in).

Now look at the use of `<<`

```ruby
>> a = { fq: [] }
=> {:fq=>[]}
>> a[:fq].object_id
=> 70309852606160
>> b = a
=> {:fq=>[]}
>> b[:fq] << 1
=> [1]
>> b[:fq] << 2
=> [1, 2]
>> b[:fq] << 3
=> [1, 2, 3]
>> b[:fq].object_id
=> 70309852606160
>> a[:fq].object_id
=> 70309852606160
```

By using `<<` we preserve the original array for `a[:fq]`.

app/models/curate_oai_provider.rb
```diff
@@ -0,0 +1,14 @@
+class CurateOaiProvider < OAI::Provider::Base
+  repository_name "CurateOaiProvider"
+  repository_url Curate.configuration.application_root_url
+  record_prefix "oai"
+  admin_email ENV.fetch('CURATE_HELP_NOTIFICATION_RECIPIENT')
+  sample_id "und:123456789"
+  register_format(CurationConcernProvider::OAI::Provider::Metadata::DublinCore.instance)
```
jeremyf, line 7: Should this be `register_format(::OAI::Provider::Metadata::DublinCore.instance)`; I'm not seeing an OAI module/class within CurationConcernProvider.

laritakr: Actually, this register_format is entirely unnecessary, as DublinCore is registered within the gem's OAI::Provider::Base class, which our CurateOaiProvider inherits from. I'll remove this line.

PR #931, opened 2020-08-19
Remove markdown

laritakr:
Removes markdown from title and description for OAI response. 

Title and description in curate allow the use of markdown. This needs to be removed from the OAI response in order for Primo to harvest the records.


app/models/curate_oai_provider/curation_concern_provider.rb
```diff
@@ -77,6 +77,8 @@ def format_response_terms(record)
       response_object = record.standardize
       response_object[:timestamp] = response_object[:date_modified].to_time
       response_object[:source] = File.join(Rails.configuration.application_root_url, 'show', record.noid)
+      response_object[:description] = strip_markdown(response_object[:description]) unless response_object[:description].nil?
```
dbrower, line 80: Add a big comment that stripping markdown on description and title is to accommodate primo.

PR #953, opened 2020-09-30
Rhel7

msuhovec:


config/environments/pre_production.rb
```diff
@@ -63,7 +63,7 @@
   # Send deprecation notices to registered listeners
   config.active_support.deprecation = :notify
 
-  config.application_root_url = "https://curatesvrpprd.library.nd.edu"
+  config.application_root_url = "https://curatesvr-prep.library.nd.edu"
```
laritakr, line 66: Why keep both pre_production and prep environment files?

laritakr (commented):
I don't understand the reasoning for both prep and pre_production environments. Could you elaborate on that a bit? Is this a temporary thing which will be cleaned up?

laritakr (approved):

PR #970, opened 2020-12-17
Handle orphan files

laritakr:
Allow administrators to view or download orphan files.
All other users receive a tombstone page instead.
Fixes CURATE-233


app/controllers/common_objects_controller.rb
```diff
@@ -21,6 +21,7 @@ def unauthorized_template
   before_filter :enforce_show_permissions, only: [:show]
   rescue_from Hydra::AccessDenied do |exception|
     respond_with curation_concern do |format|
+      format.html { render show_tombstone_page, status: 410 } if is_orphan_file? && cannot_view_orphans?
```
jeremyf, line 24: I don't know how this will resolve in relation to next line.  I would assume you'd want a clear `if…else…end` setup

app/controllers/downloads_controller.rb
```diff
@@ -185,7 +188,8 @@ def can_download?
       # However, since asset is in theory loaded as part of the before_filter, I'm relying
       # on that to not duplicate the underlying load_instance_from_solr that could happen in the
       # permission system.
-      can? :read, @asset || load_asset
+#      can? :read, @asset || load_asset
```
jeremyf, line 191: Is this intentional?

app/controllers/common_objects_controller.rb
laritakr: it seemed to work correctly when I tested it locally.

msuhovec (approved):

PR #972, opened 2021-01-26
Update Meta tags for Altmetrics

laritakr:
Use metadata tags on all work types.
Add specific tags for supporting Altmetrics.

CURATE-204


dbrower (approved):
yay! looks good.

PR #973, opened 2021-01-27
Update robots.txt

laritakr:
Google Scholar indexing requires changes to the robots.txt

Fixes CURATE-240

### DIFF
--- a/config/robots/production.txt
+++ b/config/robots/production.txt
@@ -4,3 +4,6 @@
 User-Agent: *
 Disallow: /downloads/
 Disallow: /help_requests/
+# Google Scholar indexing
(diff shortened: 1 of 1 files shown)


dbrower (approved):
I can see allowing downloads, but is it okay to also allow help_requests?

(probably? Since the bot doesn’t do posts or puts?)

laritakr:
> I can see allowing downloads, but is it okay to also allow help_requests?
> 
> (probably? Since the bot doesn’t do posts or puts?)

I don't know... that's probably up to you. That was in there before and I didn't change it what was there, just added the additional stuff the ticket specified.

PR #980, opened 2021-02-10
API changes for SFX integration

laritakr:
# CURATE-252:

- Add search by creator/author
- Add search by doi
- Add search by title

## Specifics:

- Restructure item search presenter with more generic logic.
- Searches both creator and author for given term.
- Returns both creator and author with results.
- Allow author to be added individually to fl=

# CURATE-253:

- filter by representative
- allow fl=visibility (only shows private, public, or registered)

### DIFF
--- a/app/presenters/api/items_search_presenter.rb
+++ b/app/presenters/api/items_search_presenter.rb
@@ -4,15 +4,57 @@ class Api::ItemsSearchPresenter
   include Sufia::Noid
   attr_reader :raw_response, :request_url, :documents, :pager, :query_parameters
 
-  VALID_KEYS_AND_SEARCH_FIELDNAMES = {
-    type: ["active_fedora_model_ssi", "desc_metadata__type_tesim"],
-    editor: ["edit_access_person_ssim"],
-    depositor: ["depositor_tesim"],
-    deposit_date: ["system_create_dtsi"],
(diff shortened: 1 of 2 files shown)


app/presenters/api/items_search_presenter.rb
```diff
@@ -249,3 +275,12 @@ def build_pagination_url(query_parameters)
     end
   end
 end
+
+class Array
```
jeremyf, line 279: I'm not a fan of adding a method to Array; That method will be on all Array objects.

```ruby
module ApiFieldDataTransformer
  def self.transform(field_data:, load_method:)
    if load_method == :screen_visibility
       return "public" if field_data.include?("public")
       return "registered" if field_data.include?("registered")
       return "private"
    else
       field_date.public_send(method)
    end
  end 
end
```
end

jeremyf (changes requested):

app/presenters/api/items_search_presenter.rb
```diff
@@ -249,3 +274,19 @@ def build_pagination_url(query_parameters)
     end
   end
 end
+
+module ApiFieldDataTransformer
```
jeremyf, line 278: I should've added that this may or may not make sense to be a method on the presenter.  As written it's fine

jeremyf (approved):

PR #998, opened 2021-04-16
Ability to orphan a file

laritakr:
Add a button for users with super admin rights to be able to orphan a file. Also adds email link for user with edit rights to contact with a request to remove file.

This provides an automated method to disconnect a file from a work when removal is requested, bypassing Fedora's issues with actually deleting an object. 

Completes CURATE-299.


jeremyf:
This looks reasonable…kind of curious about why we're making orphans (but do not need an explanation nor want one).

dbrower (approved):
Great.

PR #1023, opened 2021-11-15
Curate 255 filechar download bendo file directly insread of via active fedora

msuhovec:


hanstra (commented):
You've got the curate_worker_tmp dir set to /global/data/downloads but then it looks like it is hardcoded to /tmp in the application.yml. Is that what you want?

app/repository_datastreams/file_content_datastream.rb
```diff
@@ -1,19 +1,47 @@
...
   include Hydra::Derivatives::ExtractMetadata
   include Sufia::FileContent::Versions
 
   def extract_metadata
     return unless has_content?
-    # I want to run Clam first, let that possibly raise exceptions
+
+    
+    # if the content is in bendo, file_content will be an open file descriptor to the downloaded file,
+    # if not, it will be the ActiveFedora object's content datastream
```
dbrower, line 16: if `file_content` is a fd, does it need to be closed when finished?

app/repository_datastreams/file_content_datastream.rb
```diff
@@ -28,4 +56,35 @@ def characterization_runner
...
   end
 
+  def download_from_bendo
+    # Create full download path of file
+    download_file_dest = File.join(Figaro.env.curate_worker_tmpdir, File.basename(dsLocation))
+
+    begin
+      download_from_dsLocation(download_file_dest)
+    rescue Down::ServerError
+      retry if check_bendo_caching != '1'
```
dbrower, line 66: this test is effectively fine, but since there is an edge case of returning a cache status of 2 for "not-cachable" it is probably better to make it `check_bendo_caching == '0'`

Also I'm trying to figure out how tight of a loop this is. should there be a sleep in here to wait before trying again?

msuhovec: I started with a sleep(30) in there- I could put it back in....

app/repository_datastreams/file_content_datastream.rb
msuhovec: Good catch.

msuhovec:
> You've got the curate_worker_tmp dir set to /global/data/downloads but then it looks like it is hardcoded to /tmp in the application.yml. Is that what you want?

The only time the boilerplate application.yml gets used is on the developer's local box, and /tmp is always there.
In staging, prep, or production, the application.yml will be overwritten by the ansible-generated one with  the **curate_worker_tmp** value.

PR #1042, opened 2022-05-12
Updates to two gems and the schema.rb was rebuilt

rfox2:
These changes are primarily the result of getting the app to build on a new system but they are minor changes. Please review the update to the schema.rb file - that change happened when I ran the database creation commands.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -33,7 +33,7 @@ local_user: &local_user
   <<: *mysql_settings
   <<: *mysql_connection
   username: root
-  password:
+  password: password
 
 development: &development
   <<: *local_user
--- a/config/database.yml
+++ b/config/database.yml
@@ -33,7 +33,7 @@ local_user: &local_user
   <<: *mysql_settings
   <<: *mysql_connection
   username: root
-  password: password
+  password:
 
 development: &development
   <<: *local_user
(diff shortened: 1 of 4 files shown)


config/database.yml
```diff
@@ -33,7 +33,7 @@ local_user: &local_user
   <<: *mysql_settings
   <<: *mysql_connection
   username: root
-  password: password
+  password:
```
laritakr, line 36: I think this change is what caused CI to fail.

laritakr (dismissed):
Please restore the password so CI can run.

db/schema.rb
```diff
@@ -417,10 +424,6 @@
     t.string   "website",                              limit: 255
     t.string   "affiliation",                          limit: 255
     t.string   "telephone",                            limit: 255
-    t.string   "avatar_file_name",                     limit: 255
```
laritakr, line 420: These concern me a bit because I'm not sure where they came from or why they are getting removed.  I don't think we use them so it may be ok but we may want to try deploying the branch to test it before merging.

laritakr:
I think it's probably ok. Unless Tom got test deployable, you will have to
deploy to prep to test it out. But I would probably deploy the branch
before it's merged, just to be sure.

It has to do with users table which may only be affected if someone new is
added. So you might have to find someone who has never used Curate to log
in.

laritakr (approved):

PR #1059, opened 2023-08-10
sentry removal - WIP

EricDahl-NotreDame:
This PR removes all references to Sentry from the project

### DIFF
--- a/app/services/batch_ingestor.rb
+++ b/app/services/batch_ingestor.rb
@@ -117,7 +117,7 @@ class BatchIngestor
 
   def report_to_error_handler(params, code)
     exception = BatchIngestHTTPError.new("HTTP request failed with status #{code} for\n\t#{params.inspect}")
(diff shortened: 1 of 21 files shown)


app/services/batch_ingestor.rb
msuhovec, line 1: This one I'd lime to keep- it is generated when (among other events) the night Thesis/Dissertation submissions fail, and if I see it, I'll requeue the submission.

msuhovec (changes requested):
I asked to keep the Batch Ingest HTTP failure messgae- I am OK with deleting all of the others

app/services/batch_ingestor.rb
EricDahl-NotreDame: No problem

msuhovec (approved):

msuhovec (approved):

PR #1061, opened 2023-10-03
prepping site for removing submission

EricDahl-NotreDame:
- removed Login and  Contribute Your Work buttons from the site.
- added Banner text to app/views/application/_announcements.html.erb
screenshot_2023-10-03_at_1 35 15_pm_720

### DIFF
--- a/config/application.yml
+++ b/config/application.yml
@@ -41,8 +41,8 @@ AWS_ACCESS_KEY_ID: 'abcde'
 AWS_SECRET_ACCESS_KEY: 'abcde'
 bendo_api_key: more_fakery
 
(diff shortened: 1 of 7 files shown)


config/application.yml
msuhovec, line 1: This  boilerplate application.yml is overwritten at deployment time by an application.yml generated by the ansible script which deploys it. The actual file where the dates need to go is https://github.com/ndlib/dlt-ansible/blob/master/roles/curate-yml/templates/application.yml.j2  Tom or I can change it if you'd rather not.

msuhovec (approved):

msuhovec (approved):

